PR #4, opened 2019-11-21
Fix the test append the correct paths

fruch:


tests/aws/test_scylla_configure.py
```diff
@@ -10,8 +11,12 @@
 from pathlib import Path
 from http.server import HTTPServer, BaseHTTPRequestHandler
 from socketserver import ThreadingMixIn
+
+sys.path.append('../../aws')
```
bentsi: why did you need this?

fruch: cause test won't work without it
try running from the command line and you'll see the issue:
```
python3.7 -m unittest test_scylla_configure.py
```

fruch: pycharm add some of directories into the sys.path by default (like the root of the repo), the command line doesn't do that

bentsi: I mean it is enough to add `sys.path.append('../..')`

PR #8, opened 2020-02-05
Clearing up the docs on how to build from with the --localrpm option

fruch:
also added an option to download non scylla-core rpms,
so people won't need to built those every time

### DIFF
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -10,9 +10,12 @@ print_usage() {
     echo "  --repo-for-install  repository for install, specify .repo/.list file URL"
     echo "  --repo-for-update  repository for update, specify .repo/.list file URL"
     echo "  --product          scylla or scylla-enterprise"
+    echo "  --download-no-server  download all rpm needed excluding scylla from `repo-for-install`"
     exit 1
 }
 LOCALRPM=0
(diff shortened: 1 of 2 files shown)


fruch:
@eliransin after out talk today, I've update the docs here to try to make it a bit more clear

would love any input

fruch:
@bentsi, when talked with @eliransin this week about this change.
he was concerned with those type of change, in regards that now developers won't know how to built the AMI.
I've update the docs a bit in this PR, to clearify. but you might want to mention it when you present the new AMI. and point people to here to try out the instructions.

aws/ami/build_ami.sh
```diff
@@ -10,9 +10,12 @@ print_usage() {
     echo "  --repo-for-install  repository for install, specify .repo/.list file URL"
     echo "  --repo-for-update  repository for update, specify .repo/.list file URL"
     echo "  --product          scylla or scylla-enterprise"
+    echo "  --download-no-server  download all rpm needed but scylla from `repo-for-install`"
```
bentsi: except of Scylla

bentsi: excluding Scylla

fruch: fixed

PR #18, opened 2020-03-08
dist/redhat: Make RPM release independent of distribution

fruch:
To support both CentOS 7 and CentOS 8 with the same RPM, remove
distribution information from the "Release" string in the RPM spec file.

Similar fix was already done in scylla.git in commit:

scylladb/scylla@ef88e1e

Ref: scylladb/scylla#5631.


bentsi:
@fruch please work on your fork when firing PRs

PR #21, opened 2020-03-19
Add user_data_format_version tag to AMI on creation

fruch:
If `user_data_format_version` as missing scylla-cluter-tests fails to use
those AMIs with the correct user-data


fruch:
@bentsi why aren't we merging this ? until we'll address the renaming and such.
this could still prevent waste of resources, like we've seen

Issue #25, opened 2020-04-12
auto_bootstrap shouldn’t be set to false by default 


roydahan:
https://github.com/scylladb/scylla-machine-image/blob/b06535de7a4e0532f3564bb4b81a8b4da517ace3/aws/scylla_configure.py#L22

roydahan:
@bentsi , @fruch please fix it ASAP.

slivne:
please note previous ami default was true

```
commit 2319a532f6282967cc9cf7d1d24524e302b86ac0
Author: Shlomi Livne <shlomi@scylladb.com>
Date:   Sun Apr 14 09:39:21 2019 +0300

    Change the default of auto_bootstrap to be True
    
    The AMI was setting auto_bootstrap to false, allowing it to be
    ofverriden via user-metadata.
    
    Unofrtunatly, users don't follow guidlines - and we are changing the
    default to be true (and allowing it to be overriden to false) if needed
    
    Fixes: #46
    
    Signed-off-by: Shlomi Livne <shlomi@scylladb.com>
    Message-Id: <3b001e20712cc075d1b812f8b978f92fc3f807af.1555223823.git.shlomi@scylladb.com>
```

slivne:
Lets close the issue when its merged

penberg:
The `scylla-machine-image` repository uses `next` as the default branch. This needs to be switched to `master` because Github auto-closes issues as they are merged to the default branch.

bentsi:
switched the default branch to be master

PR #29, opened 2020-04-13
Change the default of auto_bootstrap to be True

fruch:
The AMI was setting auto_bootstrap to false, allowing it to be
overriden via user-metadata.

Unofourtunatly, users don't follow guidlines - and we are changing the
default to be true (and allowing it to be overriden to false) if needed

Fixes: #25

### DIFF
--- a/aws/scylla_configure.py
+++ b/aws/scylla_configure.py
@@ -19,7 +19,7 @@ class ScyllaAmiConfigurator:
         'scylla_yaml': {
             'cluster_name': "scylladb-cluster-%s" % int(time.time()),
             'experimental': False,
-            'auto_bootstrap': False,
+            'auto_bootstrap': True,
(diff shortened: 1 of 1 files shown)


fruch:
one merged, need to backported to next-4.0

bentsi:
@fruch please run tests after the fix next time:
```
11:20:34  ======================================================================
11:20:34  FAIL: test_user_data_params_are_set (test_scylla_configure.TestScyllaConfigurator)
11:20:34  ----------------------------------------------------------------------
11:20:34  Traceback (most recent call last):
11:20:34    File "/jenkins/workspace/scylla-master/next/scylla-machine-image/tests/aws/test_scylla_configure.py", line 146, in test_user_data_params_are_set
11:20:34      assert scylla_yaml["auto_bootstrap"] is False
11:20:34  AssertionError 
```

PR #30, opened 2020-04-15
fix(test): expect auto_bootstrap True

bentsi:


bentsi:
@penberg should I backport this to `next-4.0`?

penberg:
No, not until it hits `master`.

PR #33, opened 2020-05-12
dist/redhat: drop dependency on pystache

penberg:
The scylla.git toolchain moved to Fedora 32, which no longer ships
"pystache". Let's drop the dependency like scylla.git did.

### DIFF
--- a/dist/redhat/build_rpm.sh
+++ b/dist/redhat/build_rpm.sh
@@ -77,12 +77,6 @@ pkg_install python3
 pkg_install python3-devel
 pkg_install python3-pip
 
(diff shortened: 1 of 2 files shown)


fruch:
@penberg what about deleting the `spec.mustache` file ?

fruch:
@bentsi we'll need to make sure we merge this to enterprise after this get to master

penberg:
@fruch The pull request _renames_ the file.

penberg:
Queued to next as 8c2cb2d20bf34bdae33254f87fa0c3fab3c2190a, thanks @fruch!

PR #34, opened 2020-05-12
build: Make package product configurable

penberg:
This adds a PRODUCT environment variable to SCYLLA-VERSION-GEN to allow
changing the product name. This is needed for scylla-enterprise
packaging.

### DIFF
--- a/SCYLLA-VERSION-GEN
+++ b/SCYLLA-VERSION-GEN
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+PRODUCT=scylla
 VERSION=666.development
 
 if test -f version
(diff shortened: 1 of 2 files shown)


penberg:
Queued to next as 9301e93dd3e90290c9a0a707889e72638e5de1f6, thanks @fruch!

fruch:
@penberg, when merging the enterprise version, we'll need to pay attention, cause I've did some changes in this respect in enterprise.

fruch:
@penberg why not merging ? just our of habit ?

penberg:
@fruch I just use the same workflow across all repositories. What merge option do you use from the Github UI, btw? We used "squash and merge" for single-commit pull requests in scylla.git, but stopped using it because committer ended up always being "GitHub"...

fruch:
The rebase option works quite well for us.

I rarely use the squash, only if I'm trying to quickly sort out a all the
commit into one, but it tend to create messy to edit it on the UI (the
commit message)

Also the marking on squash, I think I've depends on some github
configuration, i.e. email matching to the commits email, or something like
that.

PR #39, opened 2020-05-24
Change to build with the dbuild docker

fruch:
Since releng team is going to start running the building of the RPM inside dbuild, and they encounter some issues with that.
we want to run similar image, to spot any issue beforehand.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,5 @@ jobs:
 
     - stage: building RPM
       script:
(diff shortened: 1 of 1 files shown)


bentsi:
@fruch add description: why moving to dbuild Docker?

penberg:
No objections to this, but the issues were specific to our setup, and would not have been caught by running with `dbuild`. I think the current setup is fine too.

fruch:
> No objections to this, but the issues were specific to our setup, and would not have been caught by running with `dbuild`. I think the current setup is fine too.

This PR started as trying to help yaron get rid of the `git archive` from this repo, but it was too complicated to remove Source0 from the spec file.

Using a the toolchain image is just a leftover from that, and I think we should still move to that image, to make sure we don't brake it

Maybe we could pick the one from master scylla all the time... but poorly  a "nice to have"

fruch:
this isn't needed anymore

PR #43, opened 2020-06-03
Fix AMI building straight from a repo without --localrpm

fruch:
on other prodcut (like enterprise) the $PRODUCT was used on the install command
now we pass it into the python install script


penberg:
Queued as 4fba63a0d5ffd172f01a8bb3d3f237e99fcee400.

PR #44, opened 2020-06-08
aws: switch AMI distribution to Amazon Linux 2

syuu1228:
On Amazon Linux 2, we can get more recent version kernel by default,
with EC2 device driver support.

This commit changes following things:
 - dropped AMI variable to specify source AMI, introduced source_ami_filter
  to automatically lookup latest Amazon Linux 2 AMI insteead.
 - AMI default user changes:
  - changed default user as "scyllaadm"
  - make "centos" user as an alias of "scyllaadm"
  - deleted "ec2-user" user which is default user on Amazon Linux 2
 - dropped kernel-ml installation code and udev/bootparam customize for
  EC2 support, and also initramfs generation for EC2 support. Since
  Amazon Linux 2 is optimized for EC2, we don't need these customize.
 - reuired small changes to make scripts compatible with Amazon Linux 2:
  - build time default user changed to ec2-user
  - cloud.cfg: 'ssh_deletekeys: 0' should be replaced with 'ssh_deletekeys: false'

### DIFF
--- a/aws/ami/files/scylla_install_ami
+++ b/aws/ami/files/scylla_install_ami
@@ -38,6 +38,7 @@ if __name__ == '__main__':
     if os.getuid() > 0:
         print('Requires root permission.')
         sys.exit(1)
+    homedir = os.path.abspath(os.path.join(__file__, os.pardir))
     parser = argparse.ArgumentParser(description='Construct AMI')
     parser.add_argument('--localrpm', action='store_true', default=False,
                         help='deploy locally built rpms')
@@ -63,7 +64,7 @@ if __name__ == '__main__':
         run('curl -L -o /etc/yum.repos.d/scylla_install.repo {REPO_FOR_INSTALL}'.format(REPO_FOR_INSTALL=args.repo_for_install))
 
     if args.localrpm:
-        rpms = glob.glob('/home/centos/scylla*.*.rpm')
+        rpms = glob.glob('{}/scylla*.*.rpm'.format(homedir))
         run('yum install -y {}'.format(' '.join(rpms)))
     else:
         run('yum install -y scylla-machine-image scylla-debuginfo')
@@ -84,33 +85,15 @@ if __name__ == '__main__':
     run('/opt/scylladb/scripts/scylla_setup --ntp-domain amazon --no-coredump-setup --no-sysconfig-setup --no-raid-setup --no-io-setup --no-bootparam-setup --no-ec2-check')
     run('/opt/scylladb/scripts/scylla_sysconfig_setup --ami')
     run('/opt/scylladb/scripts/scylla_bootparam_setup --ami')
-    os.remove('/home/centos/.ssh/authorized_keys')
+    os.remove('{}/.ssh/authorized_keys'.format(homedir))
     os.remove('/var/lib/scylla-housekeeping/housekeeping.uuid')
 
-    run('yum -y update kernel')
-    run('yum -y install dkms git grubby kernel-devel')
-    run('rpm -e kernel-$(uname -r)', shell=True)
-    run('rpm --import https://www.elrepo.org/RPM-GPG-KEY-elrepo.org')
-    run('rpm -Uvh http://www.elrepo.org/elrepo-release-7.0-3.el7.elrepo.noarch.rpm')
-    run('yum -y --enablerepo=elrepo-kernel install kernel-ml kernel-ml-devel')
-    os.remove('/etc/udev/rules.d/80-net-name-slot.rules')
-    with open('/etc/default/grub') as f:
-        grub = f.read()
-    grub = re.sub(r'^GRUB_CMDLINE_LINUX="(.+)"$', r'GRUB_CMDLINE_LINUX="\1 net.ifnames=0"', grub, flags=re.MULTILINE)
-    with open('/etc/default/grub', 'w') as f:
-        f.write(grub)
-    run('grub2-mkconfig -o /boot/grub2/grub.cfg')
-    c7kver = get_kver('/boot/vmlinuz-*el7.x86_64')
-    mlkver = get_kver('/boot/vmlinuz-*el7.elrepo.x86_64')
-    run('grubby --grub2 --set-default /boot/vmlinuz-{mlkver}'.format(mlkver=mlkver))
-    with open('/etc/yum.conf', 'a') as f:
-        f.write(u'exclude=kernel kernel-devel')
-
-    run('dracut --verbose --add-drivers "ixgbevf nvme ena" --force --kver {c7kver}'.format(c7kver=c7kver))
-    run('dracut --verbose --add-drivers "ixgbevf nvme ena" --force --kver {mlkver}'.format(mlkver=mlkver))
+    run('sudo amazon-linux-extras install kernel-ng')
 
     with open('/etc/cloud/cloud.cfg') as f:
         cfg = f.read()
-    cfg2 = re.sub('^ssh_deletekeys:   0', 'ssh_deletekeys:   1', cfg, flags=re.MULTILINE)
+    cfg2 = re.sub('^ssh_deletekeys:   false', 'ssh_deletekeys:   true', cfg, flags=re.MULTILINE)
+    cfg2 = re.sub('^    name: ec2-user', '    name: scyllaadm', cfg2, flags=re.MULTILINE)
     with open('/etc/cloud/cloud.cfg', 'w') as f:
         f.write(cfg2)
+    shutil.move('{}/99_user_alias.cfg'.format(homedir), '/etc/cloud/cloud.cfg.d')
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -83,9 +83,8 @@ check_rpm_exists () {
         fi
     done
 }
-AMI=ami-ae7bfdb8
 REGION=us-east-1
-SSH_USERNAME=centos
+SSH_USERNAME=ec2-user
 
 if [ $LOCALRPM -eq 1 ]; then
     INSTALL_ARGS="$INSTALL_ARGS --localrpm"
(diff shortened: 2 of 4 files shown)


fruch:
@roydahan A remind that we still need to try this one with SCT, last time I've tried it there were issues on the SCT end, probably failure in the boot script we are using in SCT.

enaydanov:
> @roydahan A remind that we still need to try this one with SCT, last time I've tried it there were issues on the SCT end, probably failure in the boot script we are using in SCT.

We have open issue on Scylla side: https://github.com/scylladb/scylla/issues/5977

avikivity:
@syuu1228 can you supply an AMI built with this patch for testing?

avikivity:
@syuu1228 let's also enable the kernel-ng repo to get a new kernel. The default one is too old (4.14, doesn't have linux-aio poll support).

slivne:
@avikivity AWS does not provide for the kernel-ng the same grantees with regards to support, please review and then verify we indeed want to switch  

https://aws.amazon.com/amazon-linux-2/faqs/

Long Term Support
Q. What is included in the Long Term Support for Amazon Linux 2?
Long-term support for Amazon Linux 2 only applies to core packages and includes:
1) AWS will provide security updates and bug fixes for all packages in core until June 30, 2023.
2) AWS will maintain user-space Application Binary Interface (ABI) compatibility for the following packages in core:

elfutils-libelf, glibc, glibc-utils, hesiod, krb5-libs, libgcc, libgomp, libstdc++, libtbb.so, libtbbmalloc.so, libtbbmalloc_proxy.so, libusb, libxml2, libxslt, pam, audit-libs, audit-libs-python, bzip2-libs, c-ares, clutter, cups-libs, cyrus-sasl-gssapi, cyrus-sasl-lib, cyrus-sasl-md5, dbus-glib, dbus-libs, elfutils-libs, expat, fuse-libs, glib2, gmp, gnutls, httpd, libICE, libSM, libX11, libXau, libXaw, libXext, libXft, libXi, libXinerama, libXpm, libXrandr, libXrender, libXt, libXtst, libacl, libaio, libatomic, libattr, libblkid, libcap-ng, libdb, libdb-cxx, libgudev1, libhugetlbfs, libnotify, libpfm, libsmbclient, libtalloc, libtdb, libtevent, libusb, libuuid, ncurses-libs, nss, nss-sysinit, numactl, openssl, p11-kit, papi, pcre, perl, perl-Digest-SHA, perl-Time-Piece, perl-libs, popt, python, python-libs, readline, realmd, ruby, scl-utils, sqlite, systemd-libs, systemtap, tcl, tcp_wrappers-libs, xz-libs, and zlib


3) AWS will provide Application Binary Interface (ABI) compatibility for all other packages in core unless providing such compatibility is not possible for reasons beyond AWS’s control.

Q. Does Amazon Linux 2 maintain kernel-space ABI compatibility?
No, Amazon Linux 2 does not maintain kernel-space ABI compatibility. If there is a change in the upstream Linux kernel that breaks ABI stability, then your applications that rely on third-party kernel drivers may require additional modifications.

Q. Does AWS backport security fixes for Amazon Linux 2?
Yes. Amazon routinely takes fixes out of the most recent version of upstream software packages and applies it to the version of the package in Amazon Linux 2. During this process, Amazon isolates the fix from any other changes, ensures that the fixes do not introduce unwanted side effects, and then applies the fixes.


**Q. Do the long-term support policies apply to extras topics?
The contents of extras topics are exempt from the Amazon Linux policy on long-term support and binary compatibility. Extras topics provide access to a curated list of rapidly evolving technologies and are likely to be updated frequently. When new versions of packages in extras topics are released, support will be provided only for the most current packages. Over time, these technologies will continue to mature and stabilize and may eventually be added to the Amazon Linux 2 "core" repositories to which the Amazon Linux 2 Long Term Support policies apply.**

(

avikivity:
We don't care about long-term support of the kernel. We are happy to move to newer versions of the kernel as they appear.

syuu1228:
> > @roydahan A remind that we still need to try this one with SCT, last time I've tried it there were issues on the SCT end, probably failure in the boot script we are using in SCT.
> 
> We have open issue on Scylla side: [scylladb/scylla#5977](https://github.com/scylladb/scylla/issues/5977)

Ah, didn't realized the issue, I will fix it.

syuu1228:
> @syuu1228 let's also enable the kernel-ng repo to get a new kernel. The default one is too old (4.14, doesn't have linux-aio poll support).

Switched to kernel-ng. https://github.com/scylladb/scylla-machine-image/pull/44/commits/e711c3d12705cbf3d338d34ac7c75531102e0780#diff-293a06c5b182b55179380ceaf0141fc0R92

syuu1228:
@avikivity Here's AMI for testing: us-east-1: ami-0395703cabcaeee08

avikivity:
@roydahan @fruch please test out the AMI that @syuu1228 provided

@syuu1228 we may need to provide an alternative to the cpupower service, but we can deal with that outside the this commit.

roydahan:
Assigned for @fgelcer for extensive testing, after that we can approve for merge.

fgelcer:
- [x] AMI test (passed)
- [x] longevity 4h (passed)
- [ ] longevity 50g 4 days (had a SCT issue, running again)
- [ ] IPv6 lonvegity (failed - investigating)
- [ ] performance tests (didn't start yet)


also, i'm seeing on all those runs an error that seems related to aws, but i'm not sure what effect it has in the cluster:
`sshd[40153]: error: AuthorizedKeysCommand /opt/aws/bin/eic_run_authorized_keys centos SHA256:g7OmDbWC8qLUZHxjMohQe3LsQHfcOJbG1ZHqodB6smc failed, status 22`

avikivity:
We need to understand that error, to make sure it won't lock out users. Looks like it's used to install keys.

It may be related to the user renaming games we are playing.

syuu1228:
@avikivity Tested the upstream AMI (Amazon Linux 2), got same error:
```
Jun 22 08:07:54 ip-10-0-0-48.ec2.internal sshd[4163]: error: AuthorizedKeysCommand /opt/aws/bin/eic_run_authorized_keys ec2-user SHA256:YyR7dXfr3tTcMcWvnC/D0J+puFWmQZXA/ZyCPcsWhjg failed, status 22
Jun 22 08:07:54 ip-10-0-0-48.ec2.internal sshd[4163]: error: AuthorizedKeysCommand /opt/aws/bin/eic_run_authorized_keys ec2-user SHA256:TBaJ0+SQuyGUVzjYQKkYYsuE1uQvUEEF6w5XZVR2mAU failed, status 22
```
Seems like this is not an issue, it is yet another way to login an EC2 instance, without manage user SSH key (it says it uses 'one-time-use SSH key')
So when we login the EC2 instance without using the method, ```AuthorizedKeysCommand /opt/aws/bin/eic_run_authorized_keys``` should always try and fail, then proceed to normal login.

reference: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Connect-using-EC2-Instance-Connect.html

----

Amazon EC2 Instance Connect provides a simple and secure way to connect to your instances using Secure Shell (SSH). With EC2 Instance Connect, you use AWS Identity and Access Management (IAM) policies and principals to control SSH access to your instances, removing the need to share and manage SSH keys. All connection requests using EC2 Instance Connect are logged to AWS CloudTrail so that you can audit connection requests.

You can use Instance Connect to connect to your Linux instances using a browser-based client, the Amazon EC2 Instance Connect CLI, or the SSH client of your choice.

When you connect to an instance using EC2 Instance Connect, the Instance Connect API pushes a one-time-use SSH public key to the instance metadata where it remains for 60 seconds. An IAM policy attached to your IAM user authorizes your IAM user to push the public key to the instance metadata. The SSH daemon uses AuthorizedKeysCommand and AuthorizedKeysCommandUser, which are configured when Instance Connect is installed, to look up the public key from the instance metadata for authentication, and connects you to the instance.

----

To use EC2 Instance Connect, we'll need to install EC2 Instance Connect CLI then execute ```mssh centos@i-INSTANCEID```, or use browser based SSH. 
reference: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-connect-methods.html

avikivity:
ok, we can live with that error.

fgelcer:
> * [x]  AMI test (passed)
> * [x]  longevity 4h (passed)
> * [x]  IPv6 lonvegity (failed - investigating)
> * [x]  longevity 50g 4 days (had a SCT issue, running again)
> * [ ]  performance tests (didn't start yet)
> 
> also, i'm seeing on all those runs an error that seems related to aws, but i'm not sure what effect it has in the cluster:
> `sshd[40153]: error: AuthorizedKeysCommand /opt/aws/bin/eic_run_authorized_keys centos SHA256:g7OmDbWC8qLUZHxjMohQe3LsQHfcOJbG1ZHqodB6smc failed, status 22`

the 50g ran for more than 24 hours and failed while running a nemesis that calls for `nodetool repair` - will try to take a look if it is a known issue, but seems unrelated to this PR.

IMHO this change can be merged, as we have now only performance tests in front of us

fruch:
@fgelcer the PR in SCT was merged ? cause one we merge this, in few hours the master would produce an AMI with it

fgelcer:
> @fgelcer the PR in SCT was merged ? cause one we merge this, in few hours the master would produce an AMI with it

not merged, but PR is ready to be merged, fixed all comments that came out from @bentsi and @enaydanov

roydahan:
@fgelcer did you test if we still need the workaround of "config_ipv6_as_persistent"?
If not, we will be able to close https://github.com/scylladb/scylla/issues/5179

fgelcer:
> @fgelcer did you test if we still need the workaround of "config_ipv6_as_persistent"?
> If not, we will be able to close [scylladb/scylla#5179](https://github.com/scylladb/scylla/issues/5179)

i have removed the `config_ipv6_as_persistent` and after nemesis `HardRebootNode` run on node-3, it got disconnected from the cluster, making the test to fail, as the node "didn't return".
after that, i connected to that node, and ran the `sudo ip route add {ipv6_cidr} dev eth0` command manually, and then it returned to the cluster...
so we can't remove this function yet.

roydahan:
ok, so leave it as is and we leave scylladb/scylla#5179 open for now.

fgelcer:
> @fgelcer did you test if we still need the workaround of "config_ipv6_as_persistent"?
> If not, we will be able to close [scylladb/scylla#5179](https://github.com/scylladb/scylla/issues/5179)

So, after re-working around the workaround, i confirm that this AMI works fine without the IPv6 workaround we added to SCT. Now we will need to fix the loaders and monitor AMIs

fgelcer:
> @fgelcer the PR in SCT was merged ? cause one we merge this, in few hours the master would produce an AMI with it

The SCT PR is merged. The perf results seems to be fine (per @roydahan 's review) and we are ready to merge this one.
@avikivity , please confirm.

aws/ami/files/scylla_install_ami
```diff
@@ -83,34 +84,17 @@ if __name__ == '__main__':
                 f.write(l)
     run('/opt/scylladb/scripts/scylla_setup --ntp-domain amazon --no-coredump-setup --no-sysconfig-setup --no-raid-setup --no-io-setup --no-bootparam-setup --no-ec2-check')
     run('/opt/scylladb/scripts/scylla_sysconfig_setup --ami')
-    run('/opt/scylladb/scripts/scylla_bootparam_setup --ami')
-    os.remove('/home/centos/.ssh/authorized_keys')
+#XXX: temporary commented out until the script fix for Amazon Linux 2
+#    run('/opt/scylladb/scripts/scylla_bootparam_setup --ami')
```
fruch, line 88: why aren't we running `scylla_bootparam_setup`, is there a scylla issue for that ?

avikivity:
> @avikivity , please confirm.

No objections from me.

aws/ami/files/scylla_install_ami
syuu1228: @fruch There was a bug that causes an error on scylla_bootparam_setup when we run it on Amazon Linux 2, but it already fixed at https://github.com/scylladb/scylla/commit/4c9369e449623f5efd3676c8954179b7c3c952b8#diff-69301dce68daf68094775d1104243d24, so we can un-comment this now. I'll push the fix.

syuu1228: Pushed.

roydahan:
@fruch / @bentsi why haven't we merged it?

fruch:
> @fruch / @bentsi why haven't we merged it?

basically cause we forgot about it...

PR #47, opened 2020-07-29
scylla_image_setup: enable swap during AMI instance init

amoskong:
This PR enable swap in scylla_image_setup, it will be enabled during AMI instance init.
The swap is always helpful, and recommended for scylla user. We don't need to make it configurable, and just enable it all the time.

Current default root disk size (10G) might be not enough if the instance type is very large. Let's increase it to 30G.

/CC @roydahan @tzach @bentsi @fruch

### DIFF
--- a/aws/ami/scylla.json
+++ b/aws/ami/scylla.json
@@ -45,7 +45,7 @@
         {
           "delete_on_termination": true,
           "device_name": "/dev/xvda",
-          "volume_size": 10
+          "volume_size": 30
         }
       ],
       "region": "{{user `region`}}",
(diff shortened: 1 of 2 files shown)


roydahan:
I don't have permissions to assign reviewers, so @fruch please review and merge.
Then, when it pass ami-test in master, let's backport to 4.2.

fruch:
@roydahan seems like we mistakly merged this one into master

Can you revert it from master, also seems like it not working...

amoskong:
> @roydahan seems like we mistakly merged this one into master
> 
> Can you revert it from master, also seems like it not working...
> [scylladb/scylla#6973 (comment)](https://github.com/scylladb/scylla/issues/6973#issuecomment-668546455)

I just created a new PR for next:
- https://github.com/scylladb/scylla-machine-image/pull/48

amoskong:
> @roydahan seems like we mistakly merged this one into master
> 
> Can you revert it from master, also seems like it not working...
> [scylladb/scylla#6973 (comment)](https://github.com/scylladb/scylla/issues/6973#issuecomment-668546455)

Hi @roydahan 

Pekka had rebased next (contained same fixes as this PR) to master. SO we don't need to revert this PR now.

image

PR #50, opened 2020-08-11
k8s: Added script for setting up k8s nodes

zimnx:
This script allows to configure k8s node disks and tune networking. 
Priviledged pods should run it as a entrypoint, and provide two environmental variables:
* ROOT_DISK - where Scylla root directory should be mounted
* ROOT_MOUNTPOINT - path to directory where host filesystem is mounted

### DIFF
--- a/aws/scylla_create_devices
+++ b/aws/scylla_create_devices
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import argparse
 import re
 import os
 import sys
@@ -26,8 +27,7 @@ from pathlib import Path
 
 raid_script = "/opt/scylladb/scripts/scylla_raid_setup"
 raid_device = "/dev/md%d"
-scylla_root = "/var/lib/scylla"
-
+scylla_root = ""
 
 def scylla_directory(role):
     if role == "all":
@@ -183,6 +183,12 @@ def get_disk_bundles():
         config_array(typemap[t], role[t], mdidx)
         mdidx += 1
 
-
 if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Disk creation script for Scylla.')
+    parser.add_argument('--scylla-data-root', dest='scylla_data_root', action='store',
+                        help='location of Scylla root data directory', default="/var/lib/scylla")
+    args = parser.parse_args()
+
+    scylla_root = args.scylla_data_root
+
     get_disk_bundles()
--- a/k8s/Dockerfile
+++ b/k8s/Dockerfile
@@ -1,0 +1,17 @@
+FROM docker.io/scylladb/scylla:4.1.6 as base
+
+# Install scripts dependencies.
+RUN yum -y install epel-release && \
+    yum -y clean expire-cache && \
+    yum -y update && \
+    yum install -y hwloc ethtool python3-yaml python3 python3-devel gcc && \
+    yum clean all
+
+RUN pip3 install pyyaml psutil
+
+ARG cloud_provider
+
+COPY $cloud_provider/scylla_create_devices /opt/scylladb/scylla-machine-image/scylla_create_devices
+COPY k8s/scylla_k8s_node_setup /opt/scylladb/scylla-machine-image/scylla_k8s_node_setup
+
+ENTRYPOINT ["/opt/scylladb/scylla-machine-image/scylla_k8s_node_setup"]
--- a/k8s/build_image.sh
+++ b/k8s/build_image.sh
@@ -1,0 +1,43 @@
+#!/bin/bash -e
+#
+# Copyright 2020 ScyllaDB
+#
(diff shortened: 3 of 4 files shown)


zimnx:
cc. @mmatczuk

k8s/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM centos:7 as base
```
fruch, line 1: is this image is going to be uploaded to dockerhub ?

fruch:
@zimnx just for some context, there was some mail thread (or any other form of agreement, with @slivne or @avikivity) that this belongs here ?

aws/scylla_create_devices
```diff
@@ -23,16 +24,14 @@ import subprocess
...
 
-setup_logging()
 logger = logging.getLogger(__name__)
 
 
 raid_script = "/opt/scylladb/scripts/scylla_raid_setup"
-raid_device = "/dev/md%d"
-scylla_root = "/var/lib/scylla"
+raid_device = ""
+scylla_root = ""
```
avikivity, line 35: Why are you changing code in the aws/ directory?

avikivity: If you think it is useful for non-aws, extract it to a library and use it from aws/k8s.

k8s/Dockerfile
avikivity, line 1: Please use centos:8 as a base.

avikivity: Also, disambiguate the repo: docker.io/centos:8.

fruch: i.e. you have plan on how to publish this docker image ? versions and such ?

aws/scylla_create_devices
zimnx: Currently only AWS is supported. It's better to have scripts parameterized than relying on hardcoded values. These values doesn't work for k8s image so I added parametrization.

k8s/Dockerfile
zimnx: I haven't thought about it, I decided to use what you guys use which is output of `SCYLLA-VERSION-GEN`. Check build_image script for details.

zimnx (commented):
Because there is a PR which adds GCE support, and script for disk formatting is different for each cloud provider I added `build_image` scripts which chooses right one depending on parameter. This means image for GCE and AWS will be different. 

In my opinion there should be single generic script which handles disks, but I leave decision up to you. In scylla operator we can live with current approach.

zimnx:
> @zimnx just for some context, there was some mail thread (or any other form of agreement, with @slivne or @avikivity) that this belongs here ?

This is needed for scylla-operator example for EKS, where we want to configure node using privileged container running as a DeamonSet.

k8s/build_image.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [[ ! -e k8s/build_image.sh ]]; then
+    echo "run build_image.sh in top of scylla-machine-image dir"
+    exit 1
+fi
+
+echo "Building in $PWD..."
+
+VERSION=$(./SCYLLA-VERSION-GEN)
+PACKAGE_NAME="scylladb/scylla-machine-image-k8s-$CLOUD_PROVIDER:$VERSION"
```
fruch, line 41: @avikivity we have a dockerhub project for that already ?

aws/scylla_create_devices
```diff
@@ -189,6 +188,16 @@ def get_disk_bundles():
         config_array(typemap[t], role[t], mdidx)
         mdidx += 1
 
-
 if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Disk creation script for Scylla.')
+    parser.add_argument('--raid-device', dest='raid_device', action='store',
+                        help='location of RAID device', default="/dev/md%d")
```
fruch, line 194: "/dev/md%d" ? who is going to use this formatting ?

zimnx: removed this param

zimnx:
should I point this PR to `next` branch or `master`?

fruch:
> should I point this PR to `next` branch or `master`?

Should point it to next

aws/scylla_create_devices
fruch: that's fine, this script is per cloud, and we now have one for GCE as well...

k8s/build_image.sh
fruch: @zimnx is this image is going to be public ? so we should open up a dockerhub project for it (under scylladb)
and talk with relang team on making sure this one is build and push as part of master, and part of formal releases.
FYI, @yaronkaikov @penberg

k8s/Dockerfile
fruch: After we merged that should be thought about...

fruch (approved):
LGTM,
but please to talk with releng team about how this docker should be versioned/published and so on... (that outside of this repo scope)

PR #57, opened 2020-09-29
fix(ena_support): move to newer centos base image

fruch:
Recently (24-25/Sep), our images started missing ENA support.
seem like out base image was very old moving to
"CentOS Linux 7 x86_64 HVM EBS ENA 1905" (ami-0e4bba886a574c2d8) 
has fixed the issue.

Fixes: scylladb/scylla#7295

### DIFF
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -83,7 +83,7 @@ check_rpm_exists () {
         fi
     done
 }
-AMI=ami-ae7bfdb8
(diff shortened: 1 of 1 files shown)


slivne:
https://aws.amazon.com/premiumsupport/knowledge-center/install-ena-driver-rhel-ec2/

avikivity (approved):

fruch:
Once it's passing next, I'll do same for 4.1

PR #64, opened 2020-10-13
Force gce image build when already exists

yaronkaikov:
in case we want to re-build gce image , it will fail by default. Adding -force flag to override it

This change required by https://github.com/scylladb/scylla-pkg/pull/1624


bentsi (approved):
LGTM

fruch (approved):
LGTM, 
but we should be extra careful with it, since in AWS we can't overwrite an image.
that we won't find ourselves overwritten formal ones that easily

yaronkaikov:
@fruch I think we should also generate the name of the image based on time or something, so every run will be with unique name

PR #65, opened 2020-10-14
Fix path when building with --repo for gce and ami 

yaronkaikov:


fruch:
you should be using `--localrpm` flag, and not changing this

when running without `--localrpm` it download them first to a place of choosing on it's own, and check their version for adding the tags, they aren't going to be used

yaronkaikov:
> you should be using `--localrpm` flag, and not changing this
> 
> when running without `--localrpm` it download them first to a place of choosing on it's own, and check their version for adding the tags, they aren't going to be used

Why do we need it? if we have a repo , why not downloading it and use it?

fruch (approved):
LGTM

PR #66, opened 2020-10-18
build_image.sh: get RPM version without downloading packages

yaronkaikov:
Today before building gce image, we need to get from repo the relevant packages version, in order to do it we are actually downloading the package although we are not using it

Setting the option to get package version information without downloading the packages

### DIFF
--- a/gce/image/build_image.sh
+++ b/gce/image/build_image.sh
@@ -65,11 +65,17 @@ while [ $# -gt 0 ]; do
     esac
 done
 
-get_version_from_rpm () {
+get_version_from_local_rpm () {
+  RPM=$1
+  RELEASE=$(rpm -qi $RPM | awk '/Release/ { print $3 }' )
+  VERSION=$(rpm -qi $RPM | awk '/Version/ { print $3 }' )
+  echo "$VERSION-$RELEASE"
+}
+
+get_version_from_remote_rpm () {
     RPM=$1
-    RELEASE=$(rpm -qi $RPM | awk '/Release/ { print $3 }' )
-    VERSION=$(rpm -qi $RPM | awk '/Version/ { print $3 }' )
-    echo "$VERSION-$RELEASE"
+    VERSION=$(yum provides $RPM | grep "Provide.*=" | awk '{print $5}')
+    echo "$VERSION"
 }
 
 check_rpm_exists () {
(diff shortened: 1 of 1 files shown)


gce/image/build_image.sh
```diff
@@ -21,14 +21,10 @@ print_usage() {
     echo "build_image.sh --localrpm --repo [URL] --target [distribution]"
     echo "  --localrpm           deploy locally built rpms"
     echo "  --repo               repository for both install and update, specify .repo file URL"
-    echo "  --repo-for-install   repository for install, specify .repo file URL"
-    echo "  --repo-for-update    repository for update, specify .repo file URL"
     echo "  --product            scylla or scylla-enterprise"
-    echo "  --download-no-server download all rpms needed excluding scylla using .repo provided in --repo-for-install"
```
fruch, line 27: Why removing this option ? This one is for who want build his custom version with his own scylla

yaronkaikov: Can't they use --localrpm?
isn't it the same?

gce/image/build_image.sh
```diff
@@ -138,9 +109,6 @@ if [ ! -f variables.json ]; then
     exit 1
 fi
 
-cd $DIR
-mkdir -p build
```
fruch, line 142: Why removing the mkdir, we need to create the build directory, otherwise it would fail

gce/image/build_image.sh
fruch: This option help gather all the other rooms, to build only the parts you want to change

fruch: I.e. see the README for the AMI for example of usage of it

gce/image/build_image.sh
yaronkaikov: mistake, it's back now

gce/image/build_image.sh
yaronkaikov: OK, i understand the different, let me fix it

yaronkaikov: Fixed and tested locally

gce/image/build_image.sh
```diff
@@ -21,8 +21,6 @@ print_usage() {
     echo "build_image.sh --localrpm --repo [URL] --target [distribution]"
     echo "  --localrpm           deploy locally built rpms"
     echo "  --repo               repository for both install and update, specify .repo file URL"
-    echo "  --repo-for-install   repository for install, specify .repo file URL"
```
fruch, line 24: why are we removing those options ?

gce/image/build_image.sh
```diff
@@ -65,11 +65,17 @@ while [ $# -gt 0 ]; do
...
+  VERSION=$(rpm -qi $RPM | awk '/Version/ { print $3 }' )
+  echo "$VERSION-$RELEASE"
+}
+
+get_version_from_remote_rpm () {
     RPM=$1
-    RELEASE=$(rpm -qi $RPM | awk '/Release/ { print $3 }' )
-    VERSION=$(rpm -qi $RPM | awk '/Version/ { print $3 }' )
-    echo "$VERSION-$RELEASE"
+    VERSION=$(dnf provides $RPM | grep "Provide.*=" | awk '{print $5}')
```
fruch, line 77: can we switch to something move easily available, like `yum`, `dnf` only exists in fedora.
for when we want to debug, or run this manually

yaronkaikov: yes, it will give the same results

gce/image/build_image.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+yum#!/bin/bash -e
```
fruch, line 1: I think this is a typo...

fruch (approved):
LGTM

PR #67, opened 2020-10-19
build_ami.sh: get RPM version without downloading packages

yaronkaikov:
Today before building AMI image, we need to get from repo the relevant packages version,
In order to do it we are actually downloading the package although we are not using it

Setting the option to get package version information without downloading the packages

based on fix commit https://github.com/scylladb/scylla-machine-image/commit/56e94c5252e6da37928c329c1b0a9e95b22e853b


fruch (approved):
LGTM

PR #68, opened 2020-10-20
get_version_from_remote_rpm: get latest version on release

yaronkaikov:
An issue was raised by Takuya in https://github.com/scylladb/scylla/issues/7465

upon release when you run get_version_from_remote_rpm, you get an output for all releases instead off the latest one only

Fixing it here


fruch (approved):
LGTM

PR #69, opened 2020-10-20
aws: configure clocksource on instance first boot time

syuu1228:
To use kvm-clock as clocksource on AWS Nitro VMs, run scylla_bootparam_setup on
instance first boot time, instead of AMI image creation time.

see: scylladb/scylla#7444


syuu1228:
This is part of https://github.com/scylladb/scylla/pull/7468

penberg:
Looks good to me, @avikivity can you also please have a look?

syuu1228:
I think I need to close this due to https://github.com/scylladb/scylla/pull/7468#issuecomment-714575429, I'll make another PR

PR #77, opened 2020-11-01
improvement(scylla_create_devices): remove -c option from build_rpm script and make create devices script general

bentsi:
changes
- removed "-c" flag from build_rpm script
- generalized _scylla_create_devices_ to be able to run on GCE and AWS from the same script
- remove cloud formation copy since it is not related to build_rpm process


dist/redhat/build_rpm.sh
```diff
@@ -105,15 +101,11 @@ parameters=(
...
     -D"scylla true"
 )
 
 if [[ "$TARGET" = "centos7" ]] || [[ "$TARGET" = "centos8" ]]; then
     rpmbuild "${parameters[@]}" -ba --define '_binary_payload w2.xzdio' --define "_topdir $RPMBUILD" --define "dist .el7" $RPM_JOBS_OPTS $RPMBUILD/SPECS/$PACKAGE_NAME.spec
 else
     rpmbuild "${parameters[@]}" -ba --define '_binary_payload w2.xzdio' --define "_topdir $RPMBUILD" $RPM_JOBS_OPTS $RPMBUILD/SPECS/$PACKAGE_NAME.spec
-fi
-if [[ "$CLOUD_PROVIDER" = "aws" ]]; then
-  cp ${RPMBUILD}/../aws/cloudformation/scylla.yaml $RPMBUILD/CLOUDFORMATION/scylla_cluster_${SCYLLA_VERSION}_${SCYLLA_RELEASE}.yaml
```
yaronkaikov, line 118: Why removing cloudformation.yaml ? it's needed for AWS , right?

yaronkaikov (changes requested):

dist/redhat/build_rpm.sh
bentsi: the file `aws/cloudformation/scylla.yaml` is constant and doesn't change

fruch: It's part of the release, we have a matching on for each release. 

It wasn't changing so far, but it might.

We should leave it in, that what agreed with @tzach and @slivne.

yaronkaikov: @fruch Since this yaml file is something that you don't generating during rpm build process (except adding the scylla version) it doesn't really matter where do you do it. don't mind that it will stay as is, but we can also do it as part of the pipeline

bentsi: @fruch the script is called build_rpm.sh. Its purpose to **build RPM** and not do anything else. 
In addition, per our discussion with Yaron (and as agreed today on the release status meeting with Shlomi and Avi) the script should be cloud provider agnostic, so Yaron will implement this part in the pipeline.

yaronkaikov:
Testing the changes:

bentsi:
changes verified in: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/releng-artifacts/job/artifacts-gce-image-test/5/console
(test failed due to an issue in Scylla that is fixed here: https://github.com/scylladb/scylla/pull/7524)
We can merge this. @yarongilor please approve

yaronkaikov (approved):

PR #78, opened 2020-11-02
fix(gce image): do not setup swap when creating an image

bentsi:
scylla_swap_setup runs as part of scylla_setup during first boot of the instance that is created from the image


yaronkaikov (approved):
LGTM

PR #79, opened 2020-11-02
improvement(version): 666.development is renamed to x.y.dev

bentsi:


SCYLLA-VERSION-GEN
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 PRODUCT=scylla
-VERSION=666.development
+VERSION=x.y.dev
```
fruch, line 4: this would break the install of the RPM, this one need to match what's on master exactly.

fruch (changes requested):
this change need to happen in tandem with the change in master. otherwise it won't work

SCYLLA-VERSION-GEN
yaronkaikov: Why are we using x.y.dev instead of 4.4.dev?  the change first need to be merged in Scylla.

bentsi: updated to 4.4.dev

bentsi:
> this change need to happen in tandem with the change in master. otherwise it won't work

I know, this is the "preparation" PR

yaronkaikov:
@bhalevy will send a patch do the same change in Scylla

yaronkaikov (approved):

yaronkaikov:
https://github.com/scylladb/scylla/pull/7541 Scylla PR was merged

yaronkaikov:
@bentsi Please merge it

bentsi:
merged.

PR #82, opened 2020-11-17
GCE image:Set a uniuqe image name

yaronkaikov:
Today the GCE image name is based on product=scylla and rpm version.
When we are building GCE image more then once based on the same rpm version the image is overriding and deleted

Setting a unique image name so we can keep the history of images

Backport to 4.3 is needed

### DIFF
--- a/gce/image/build_image.sh
+++ b/gce/image/build_image.sh
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 PRODUCT=scylla
+BUILD_ID=$(date -u '+%FT%H-%M-%S')
 DIR=$(dirname $(readlink -f $0))
 
 print_usage() {
@@ -26,6 +27,7 @@ print_usage() {
     echo "  --product            scylla or scylla-enterprise"
     echo "  --download-no-server download all rpms needed excluding scylla using .repo provided in --repo-for-install"
     echo "  --dry-run            validate template only (image is not built)"
+    echo "  --build-id           Set unique build ID, will be part of GCE image name"
     exit 1
 }
 LOCALRPM=0
@@ -54,6 +56,10 @@ while [ $# -gt 0 ]; do
             ;;
         "--product")
             PRODUCT=$2
+            shift 2
+            ;;
+        "--build-id")
+            BUILD_ID=$2
             shift 2
             ;;
         "--download-no-server")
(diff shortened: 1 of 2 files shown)


penberg:
@yaronkaikov @bentsi Why is `get_version_from_local_rpm` not sufficient as a unique identifier for the image?

yaronkaikov:
Because if you build twice based on same RPM it will be override

yaronkaikov:
@penberg And when running jobs in jenkins we are not using this function, we are using `get_version_from_remote_rpm`

penberg:
So how about keeping `--build-id` command line option, but defaulting to `$(date -u '+%FT%T')` (timestamp in UTC)? That way we don't need to pass the build ID from release pipelines.

yaronkaikov:
> So how about keeping `--build-id` command line option, but defaulting to `$(date -u '+%FT%T')` (timestamp in UTC)? That way we don't need to pass the build ID from release pipelines.

@penberg Good point, will set it as default

gce/image/build_image.sh
```diff
@@ -26,6 +27,7 @@ print_usage() {
     echo "  --product            scylla or scylla-enterprise"
     echo "  --download-no-server download all rpms needed excluding scylla using .repo provided in --repo-for-install"
     echo "  --dry-run            validate template only (image is not built)"
+    echo "  --build-id           jenkins job build ID, will be use for image name"
```
penberg, line 30: Please fix the description not to refer to Jenkins build ID.

yaronkaikov: Changed

bentsi:
Naming the image should be according: https://cloud.google.com/compute/docs/naming-resources#resource-name-format
```
- Be 1-63 characters in length.
- Comply with RFC 1035 conventions.
- Match the regular expression [a-z]([-a-z0-9]*[a-z0-9])?. This means that the first character must be a lowercase letter, and all the following characters must be hyphens, lowercase letters, or digits, except the last character, which cannot be a hyphen.
```

yaronkaikov:
> Naming the image should be according: https://cloud.google.com/compute/docs/naming-resources#resource-name-format
> 
> ```
> - Be 1-63 characters in length.
> - Comply with RFC 1035 conventions.
> - Match the regular expression [a-z]([-a-z0-9]*[a-z0-9])?. This means that the first character must be a lowercase letter, and all the following characters must be hyphens, lowercase letters, or digits, except the last character, which cannot be a hyphen.
> ```

@bentsi this is not good? `scylla-4-4-dev-build-2020-11-17T10:01:56`

yaronkaikov:
@bentsi so this will be ok? `2020-11-17T10-07-50`

bentsi:
yep

yaronkaikov:
Good, changing it

yaronkaikov:
@bentsi Done

bentsi (approved):

PR #83, opened 2020-11-22
Enable clocksource setting on AMI

syuu1228:
Instead of hardcoding clocksource=tsc on bootparameter, we need to configure
it ondemand, since we have different types of hypervisor which required
different clocksource to optimize.
To do so, we need to run perftune.py tune-clock mode to auto-configure clocksource.

Related scylladb/scylla#7444

### DIFF
--- a/aws/ami/files/scylla_install_ami
+++ b/aws/ami/files/scylla_install_ami
@@ -85,7 +85,7 @@ if __name__ == '__main__':
             if not re.match(r'^ - mounts\n$', l):
                 f.write(l)
     run('/opt/scylladb/scripts/scylla_setup --ntp-domain amazon --no-coredump-setup --no-sysconfig-setup --no-raid-setup --no-io-setup --no-bootparam-setup --no-ec2-check --no-swap-setup')
-    run('/opt/scylladb/scripts/scylla_sysconfig_setup --ami')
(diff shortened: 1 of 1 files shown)


fruch (approved):

bentsi (commented):
1) why do we need to set it twice, during image creation and when the instance of the image is created
2) do we need to do this in GCE image(`gce/image/files/scylla_install_image`)?

syuu1228:
> 1. why do we need to set it twice, during image creation and when the instance of the image is created
I misunderstood that 2nd time of scylla_sysconfig_setup might overwrite SET_CLOCKSOURCE parameter to "no" if we run it without --set-clocksource, but it was wrong.
The script only changes fields that specified parameters, so I dropped it from this PR.

But I become curious do we really need to run scylla_sysconfig_setup twice at all, I'm checking the code but I suppose we only need to run it once.

> 2. do we need to do this in GCE image(`gce/image/files/scylla_install_image`)?
Yes, I think it become optimal setting, since GCE is KVM based too, just like AWS newer instances.

syuu1228:
ping

penberg (approved):
lgtm, @bentsi?

syuu1228:
ping

syuu1228:
ping

fruch (approved):
LGTM

PR #84, opened 2020-11-22
switch to subprocess.run()

syuu1228:
Since 3fefa52, we dropped scylla_util.run() and scylla_util.out(), we need to
use subprocess.run() instead.

Related scylladb/scylla#7667

Fixes: scylladb/scylla/issues#7695


fruch (approved):
LGTM

PR #85, opened 2020-12-02
Revert "aws: switch AMI distribution to Amazon Linux 2"

syuu1228:
This reverts commit de1bb48faf75fc2edbfa83dc9108d69217476a21.

We have problem on Amazon Linux kernel on scylladb/scylla#7607, have to
moving back to CentOS7.

### DIFF
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -90,8 +90,9 @@ check_rpm_exists () {
         fi
     done
 }
+AMI=ami-00e87074e52e6c9f9
 REGION=us-east-1
-SSH_USERNAME=ec2-user
+SSH_USERNAME=centos
 
 if [ $LOCALRPM -eq 1 ]; then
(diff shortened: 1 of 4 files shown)


fruch:
@slivne just to clearify, we are reverting master and backport to 4.3 ?
or we want it directly/only on 4.3 ?

fruch (approved):
LGTM

fruch:
passed OSS next, I've backported to next-4.3

penberg:
@fruch We want to revert this in master, and then backport, like we always do. The plan is to move master to CentOS 8 for 4.4.

fruch:
> @fruch We want to revert this in master, and then backport, like we always do. The plan is to move master to CentOS 8 for 4.4.

this is the way (just wanted to clarify, since there were case we skipped master, cause of reasons)
@slivne already answered me last night

syuu1228:
@penberg @fruch Here's CentOS8 support for master branch: https://github.com/scylladb/scylla-machine-image/pull/87

PR #86, opened 2020-12-03
aws: drop NIC driver repo and kernel development packages

syuu1228:
Since we moved to kernel-ml, we stopped using scylla-ixgbevf/scylla-ena
because kernel-ml contains updated drivers.
So we don't need to install NIC driver repo, also don't need development
packages to build kernel module.

### DIFF
--- a/aws/ami/files/scylla_install_ami
+++ b/aws/ami/files/scylla_install_ami
@@ -61,7 +61,6 @@ if __name__ == '__main__':
 
     run('yum update -y')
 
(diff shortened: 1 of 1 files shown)


syuu1228:
Related: https://github.com/scylladb/scylla-ami/commit/45172331ed690a0a87fb4bf94e6fa51b61f58cb0

syuu1228:
Note: this should backport to 4.3, since this is for CentOS7 AMI

fruch (approved):
LGTM

PR #87, opened 2020-12-03
aws: switch AMI distribution to CentOS8

syuu1228:
Upgrade AMI distribution from CentOS7 to CentOS8.

----

Please merge https://github.com/scylladb/scylla-machine-image/pull/86 first, since this PR depends on it.
After it merged, I'll rebase this.

Note: This one is for master.

Related: #85

### DIFF
--- a/aws/ami/scylla.json
+++ b/aws/ami/scylla.json
@@ -76,6 +76,7 @@
     },
     {
       "inline": [
+        "sudo /usr/bin/dnf -y install python2",
         "sudo /home/{{user `ssh_username`}}/scylla_install_ami {{ user `install_args` }}"
       ],
       "type": "shell"
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -90,7 +90,7 @@ check_rpm_exists () {
         fi
     done
 }
-AMI=ami-00e87074e52e6c9f9
+AMI=ami-01ca03df4a6012157
 REGION=us-east-1
 SSH_USERNAME=centos
 
--- a/aws/ami/files/scylla_install_ami
+++ b/aws/ami/files/scylla_install_ami
@@ -91,7 +91,7 @@ if __name__ == '__main__':
 
     run('yum -y install grubby')
     run('rpm --import https://www.elrepo.org/RPM-GPG-KEY-elrepo.org')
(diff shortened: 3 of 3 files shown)


aws/ami/scylla.json
```diff
@@ -76,6 +76,7 @@
     },
     {
       "inline": [
+        "sudo /usr/bin/dnf -y install python2",
```
fruch, line 79: Why do we need python2  for cqlsh ?

syuu1228: scylla_install_ami is implemented in python2, since CentOS7 always has it, and only available version of python on the distribution.
It probably can work on python3 as well, but I don't find python2/3 in CentOS8 AMI, so we need to install prior to running the script.
Otherwise it just fail.
We may able to convert it to bash if python is not installed by default, though.

syuu1228:
@avikivity @slivne Can we continue to switch distribution to CentOS8? Or we need to think again?
I just saw the announce witch says CentOS project will discontinue CentOS8 by the end of 2021, users have to switch to CentOS Stream after that:
https://blog.centos.org/2020/12/future-is-centos-stream/

aws/ami/scylla.json
syuu1228: I just verified it also works with python3.

fruch: so this can go away ?

fruch:
> @avikivity @slivne Can we continue to switch distribution to CentOS8? Or we need to think again?
> I just saw the announce witch says CentOS project will discontinue CentOS8 by the end of 2021, users have to switch to CentOS Stream after that:
> https://blog.centos.org/2020/12/future-is-centos-stream/

yeah, just read that this morning. (and few other blogs referring to that),  raises a few questions

avikivity:
> @avikivity @slivne Can we continue to switch distribution to CentOS8? Or we need to think again?
> I just saw the announce witch says CentOS project will discontinue CentOS8 by the end of 2021, users have to switch to CentOS Stream after that:
> https://blog.centos.org/2020/12/future-is-centos-stream/

It's a good question.

Our options are:
 - switch to centos 8, and trust that Red Hat will not make it unstable
 - switch to centos 8, and trust that Red Hat will back down after everyone threatens to switch to Ubuntu
 - stick with centos 7
 - switch to Oracle Linux 8
 - switch to Ubuntu 20.04 LTS

avikivity:
I started a wide conversation on the mailing lists.

PR #88, opened 2020-12-10
feature(volumes_support): Support for aws instance ebs or nvme devices

aleksbykov:
Added support of creating raid and mount to /var/lib/scylla on attached
devices (ebs volumes for aws and nvme* volumes for gce) and instance_store
volumes for aws.

### DIFF
--- a/common/scylla_configure.py
+++ b/common/scylla_configure.py
@@ -45,7 +45,7 @@ class ScyllaMachineImageConfigurator:
         'developer_mode': False,
         'post_configuration_script': '',
         'post_configuration_script_timeout': 600,  # seconds
-        'start_scylla_on_first_boot': True,
+        'start_scylla_on_first_boot': True
     }
 
     DISABLE_START_FILE_PATH = Path("/etc/scylla/ami_disabled")
@@ -145,12 +145,22 @@ class ScyllaMachineImageConfigurator:
             LOGGER.info("Disabling Scylla start on first boot")
             self.DISABLE_START_FILE_PATH.touch()
 
+    def create_devices(self):
+        device_type = self.instance_user_data.get("data_device")
+        use_device_type = "--data-device {}".format(device_type) if device_type else ""
+
+        try:
+            subprocess.run("/opt/scylladb/scylla-machine-image/scylla_create_devices {}".format(use_device_type), shell=True, check=True)
+        except Exception as e:
+            LOGGER.error("Failed to create devices: %s", e)
+
     def configure(self):
         self.configure_scylla_yaml()
         self.configure_scylla_startup_args()
         self.set_developer_mode()
         self.run_post_configuration_script()
         self.start_scylla_on_first_boot()
+        self.create_devices()
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


common/scylla_configure.py
```diff
@@ -46,6 +46,8 @@ class ScyllaMachineImageConfigurator:
         'post_configuration_script': '',
         'post_configuration_script_timeout': 600,  # seconds
         'start_scylla_on_first_boot': True,
+        'data_device': {'gce': 'attached',  # attached(default) - use attached devices nvme on gce and ebs on aws
```
bentsi, line 49: this is part of the product, users of GCP shouldn't know and bother about AWS... 
so as discussed it should be:
'data_device': 'attached' or 'instance_store'
you should set the default values in the SCT

aleksbykov: I added this so user don't know difference between gce or aws. It will automatically be choosed based on instance where it running if this parameter was not passed from sct or any other tool

aleksbykov:
@fruch could you review?

common/scylla_configure.py
fruch: I agree with bentsi move this default handling to `common/scylla_create_devices`, if user didn't pass that down,
don't pass it down to `common/scylla_create_devices` and let it handle the defaults there.

fruch (changes requested):
move default per cloud to `common/scylla_create_devices`

fruch:
@aleksbykov did you build AWS and GCP images ? and tested this on both ?

aleksbykov:
> @aleksbykov did you build AWS and GCP images ? and tested this on both ?

Could you suggest the safe way how to do that?
I only tested by manually replace the py scripts in instances and running the scylla_image_setup script manually?

fruch:
> > @aleksbykov did you build AWS and GCP images ? and tested this on both ?
> 
> Could you suggest the safe way how to do that?
> I only tested by manually replace the py scripts in instances and running the scylla_image_setup script manually?

don't know how safe it is :), but it might need a few fixes in those instructions, but general should be the way to test build locally an AMI:
https://github.com/scylladb/scylla-machine-image/blob/master/aws/ami/README.md#build-ami-from-locally-built-rpm

fruch (approved):
LGTM,
but I would merge only after you build yourself an AMI, and tested it with https://github.com/scylladb/scylla-cluster-tests/pull/2984

fruch:
@aleksbykov also if during the process you have comment/notes for the README, on how to get AMI built locally. put in in a separate commit and we'll merge it together

aleksbykov:
@fruch could review.
I added instructions to README how to build the ami from localhost.
Built ami: ami-0acea99e4341f17de
It is succesfully used in next staging job:
- regular sct job without ebs volumes:
  https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-1tb-7days-2Disruptive-Nemesises/80
- perf jobs with attached ebs volumes:
  - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/perf-branch-staging-latency/39/
 - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/scylla-master-perf-regression-aws-test-throughput/49/

fruch (commented):
LGTM

PR #89, opened 2020-12-14
aws: support building aarch64 AMI

syuu1228:
Currently we don't provide aarch64 rpm on our official repo, but it can
build locally by standard procedure of building package, on aarch64 machine.
Let's support the architecture on scylla-machine-image too, for testing purposes.

Related: scylladb/scylla#7026

### DIFF
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -81,7 +81,7 @@ get_version_from_remote_rpm () {
 
 check_rpm_exists () {
     BASE_DIR=$1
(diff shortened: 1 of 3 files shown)


penberg (approved):
lgtm, @bentsi?

fruch:
@syuu1228 @penberg  we have such RPMs readily available ? also did you manually built such AMI ?

fruch (approved):
LGTM

fruch:
also if we merge this #87 will need to change, @slivne @penberg, are we still moving to centos8 ? if we do, I would rather merged #87 first

syuu1228:
About switching CentOS8, I think we need more time to decide what to do, so this can merge first (it's CentOS7 based)

PR #90, opened 2021-01-04
Set product for AMI and GCE based on SCYLLA-VERSION-GEN

yaronkaikov:
Today we have PRODUCT definition in SCYLLA-VERSION-GEN , let's start
using it as a source instead of defining it manually for both AMI and
GCE. This way we can have a generic code for both open source and
enterprise

### DIFF
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -14,7 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-PRODUCT=scylla
+source ../../SCYLLA-VERSION-GEN
+
+PRODUCT=$(cat build/SCYLLA-PRODUCT-FILE)
 DIR=$(dirname $(readlink -f $0))
 
 print_usage() {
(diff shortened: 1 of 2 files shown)


yaronkaikov:
Testing: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/194/console

aws/ami/build_ami.sh
```diff
@@ -14,7 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-PRODUCT=scylla
+PRODUCT=$(cat ../../SCYLLA-VERSION-GEN | grep PRODUCT= | sed -e 's/PRODUCT=//')
```
fruch, line 17: why not run `SCYLLA-VERSION-GEN` ? 
and use `build/SCYLLA-PRODUCT-FILE` ?

also why not just, 
`source ../../SCYLLA-VERSION-GEN` 
?

yaronkaikov: You are right, changed

fruch (approved):
LGTM

Issue #91, opened 2021-01-11
artifacts-gce-image-test: Failed to create devices


juliayakovlev:
**System under test**

Actual gce image used: scylla-4-4-dev-build-2021-01-11t09-05-45
AWS region: us-east1
Instance type: n1-standard-2
Scylla packages installed:
scylla-conf-4.4.dev-0.20210111.3aa80f47fec.x86_64
scylla-kernel-conf-4.4.dev-0.20210111.3aa80f47fec.x86_64
scylla-tools-4.4.dev-0.20210111.3aa80f47fec.noarch
scylla-debuginfo-4.4.dev-0.20210111.3aa80f47fec.x86_64
scylla-python3-4.4.dev-0.20210111.3aa80f47fec.x86_64
scylla-node-exporter-4.4.dev-0.20210111.3aa80f47fec.x86_64
scylla-tools-core-4.4.dev-0.20210111.3aa80f47fec.noarch
scylla-jmx-4.4.dev-0.20210111.3aa80f47fec.noarch
scylla-machine-image-4.4.dev-20210111.b23ecb53ef.noarch
scylla-server-4.4.dev-0.20210111.3aa80f47fec.x86_64
scylla-4.4.dev-0.20210111.3aa80f47fec.x86_64

[Job](https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-gce-image-test/59)

Test: artifacts-gce-image-test

Failed to create devices:
```
Jan 11 09:21:07 artifacts-gce-image-jenkins-db-node-d0e9db1d-0-1 scylla_image_setup[744]: nvme0n1 is not found
Jan 11 09:21:07 artifacts-gce-image-jenkins-db-node-d0e9db1d-0-1 scylla_image_setup[744]: Traceback (most recent call last):
Jan 11 09:21:07 artifacts-gce-image-jenkins-db-node-d0e9db1d-0-1 scylla_image_setup[744]:   File "/opt/scylladb/scylla-machine-image/libexec/scylla_create_devices", line 74, in <module>
Jan 11 09:21:07 artifacts-gce-image-jenkins-db-node-d0e9db1d-0-1 scylla_image_setup[744]:     create_raid(disk_devices)
Jan 11 09:21:07 artifacts-gce-image-jenkins-db-node-d0e9db1d-0-1 scylla_image_setup[744]:   File "/opt/scylladb/scylla-machine-image/libexec/scylla_create_devices", line 33, in create_raid
[...]

PR #92, opened 2021-01-11
fix(gce_device_path): fix get full device path for gce image

aleksbykov:
method from scylla_util for gce instance return only name
of empherial device. Fix method to get_default_devices full path
for gce instances

Fix issue: https://github.com/scylladb/scylla-machine-image/issues/91

### DIFF
--- a/common/scylla_create_devices
+++ b/common/scylla_create_devices
@@ -49,12 +49,14 @@ def get_disk_devices(instance, device_type):
 
 
 def get_default_devices(instance):
+    disk_names = []
     if is_ec2():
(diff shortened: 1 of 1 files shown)


aleksbykov:
Building ami to check fix

aleksbykov:
@fruch could you review

fruch:
> Building ami to check fix

FYI, the issue was with the GCP image...

fruch (approved):
LGTM, I'll merge once you confirm you tested it

aleksbykov:
Tested for aws ami. Worked.
Finding a way to build custom gce image

PR #97, opened 2021-02-08
aws: add support package manifest file

syuu1228:
To fullfill SOC2 requirement, we need to make sure OS is up to date.
However, just running 'yum update' on whole cluster nodes may install
different versions of packages, that can cause issue on Scylla.
So generate package manifest that records package versions of all
required packages for Scylla, and use the manifest for Scylla upgrade.

Package manifest build while generating AMI, download to
build/scylla-packages.txt on build host.

Related with https://github.com/scylladb/scylla-pkg/issues/1869

### DIFF
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -148,6 +148,12 @@
         "sudo /home/{{user `ssh_username`}}/scylla_install_image --target-cloud {{build_name}} {{user `install_args`}}"
       ],
       "type": "shell"
+    },
+    {
+      "source": "/home/{{user `ssh_username`}}/scylla-packages.txt",
+      "destination": "build/",
+      "direction": "download",
+      "type": "file"
     }
   ],
   "variables": {
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -245,3 +245,20 @@ if __name__ == '__main__':
                 f.write(u'exclude=kernel kernel-devel')
         elif platform.machine() == 'aarch64':
             run('yum -y --enablerepo=centos-kernel update kernel')
+
+    # generate package manifest to scylla-packages.txt
+    if distro == 'centos':
+        run('yum install -y yum-utils')
+        packages = subprocess.check_output('repoquery --requires --resolve --recursive --installed scylla', stderr=subprocess.STDOUT, shell=True)
+        with open('{}/scylla-packages.txt'.format(homedir), 'w') as f:
+            f.write(packages)
+    else:
+        deps = subprocess.check_output('apt-cache depends --recurse --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances --installed scylla', stderr=subprocess.STDOUT, shell=True, encoding='utf-8').splitlines()
+        pkgs=[]
+        for d in deps:
+            if re.match(r'^\w', d) and not re.match(r'.+:i386$', d) and d not in pkgs:
+                pkgs.append(d)
+        with open('{}/scylla-packages.txt'.format(homedir), 'w') as f:
+            for pkg_name in pkgs:
+                pkg_name_version = subprocess.check_output("dpkg-query --showformat='${{Package}}-${{Version}}' --show {}".format(pkg_name), shell=True, encoding='utf-8')
+                f.write('{}\n'.format(pkg_name_version))


penberg:
@avikivity Your input is needed for this. It's one component in providing a list of packages to upgrade when using an existing AMI.

aws/ami/files/scylla_install_ami
```diff
@@ -121,3 +121,8 @@ if __name__ == '__main__':
             f.write(u'exclude=kernel kernel-devel')
     elif platform.machine() == 'aarch64':
         run('yum -y --enablerepo=centos-kernel update kernel')
+
+    run('yum install -y yum-utils')
+    packages = subprocess.check_output('repoquery --requires --resolve --recursive scylla', stderr=subprocess.STDOUT, shell=True)
+    with open('{}/packages.txt'.format(homedir), 'w') as f:
```
penberg, line 127: Perhaps `scylla-packages.txt` for the name?  @avikivity any preference on the location of this file?

syuu1228: Renamed.

syuu1228:
ping

avikivity:
What is this about? What's the manifest file, and why do we need it?

aws/ami/files/scylla_install_ami
```diff
@@ -121,3 +121,8 @@ if __name__ == '__main__':
             f.write(u'exclude=kernel kernel-devel')
     elif platform.machine() == 'aarch64':
         run('yum -y --enablerepo=centos-kernel update kernel')
+
+    run('yum install -y yum-utils')
+    packages = subprocess.check_output('repoquery --requires --resolve --recursive scylla', stderr=subprocess.STDOUT, shell=True)
```
avikivity, line 126: We could also do `rpm -qa`, but maybe you need this repoquery for some reason. I can't tell because I don't know what the requirements are.

syuu1228:
Fixed commit message to describe more detail, added Ubuntu AMI support.

aws/ami/files/scylla_install_ami
```diff
@@ -167,3 +167,22 @@ if __name__ == '__main__':
                 f.write(u'exclude=kernel kernel-devel')
         elif platform.machine() == 'aarch64':
             run('yum -y --enablerepo=centos-kernel update kernel')
+
+    if distro == 'centos':
```
avikivity, line 171: This will fail on RHEL and other derivatives.  We always created the AMI in the past on CentOS, but better to be prepared.

avikivity:
Thanks for explaining. It's okay to have it here, but maybe it should be a more generic tool. Anyway we can start with the way it is now; just see the minor comment above.

aws/ami/files/scylla_install_ami
syuu1228: Well, distribution detection code on scylla_install_ami is much simplified than scylla setup scripts, it just checking existance of /etc/redhat-release, and detect like this:
if exists -> 'centos' (= RHEL based distro) / if not exists -> 'ubuntu' (= non-RHEL distro).
So it actually works with RHEL & OEL.

To get more specific distribution information we need to parse /etc/os-release, but I think this is enough for current implementation.

aws/ami/files/scylla_install_ami
```diff
@@ -167,3 +167,22 @@ if __name__ == '__main__':
...
+        run('yum install -y yum-utils')
+        packages = subprocess.check_output('repoquery --requires --resolve --recursive scylla', stderr=subprocess.STDOUT, shell=True)
+        with open('{}/scylla-packages.txt'.format(homedir), 'w') as f:
+            f.write(packages)
+    else:
+        deps = subprocess.check_output('apt-cache depends --recurse --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances --installed scylla', stderr=subprocess.STDOUT, shell=True, encoding='utf-8').splitlines()
+        pkgs=[]
+        for d in deps:
+            if re.match(r'^\w', d) and not re.match(r'.+:i386$', d) and d not in pkgs:
```
bentsi, line 180: just wondering, why do you filter out i386 packages? (and not filter el7.i686 for CentOS?)

aws/ami/files/scylla_install_ami
```diff
@@ -167,3 +167,22 @@ if __name__ == '__main__':
...
+                pkgs.append(d)
+        with open('{}/scylla-packages.txt'.format(homedir), 'w') as f:
+            for p in pkgs:
+                status = subprocess.check_output('dpkg -s {}'.format(p), shell=True, encoding='utf-8').splitlines()
```
bentsi, line 184:
```suggestion
                status = subprocess.check_output("dpkg-query --showformat='${Package}-${Version}' --show {}".format(p), shell=True, encoding='utf-8').splitlines()
```

aws/ami/files/scylla_install_ami
```diff
@@ -167,3 +167,22 @@ if __name__ == '__main__':
...
+        with open('{}/scylla-packages.txt'.format(homedir), 'w') as f:
+            f.write(packages)
+    else:
+        deps = subprocess.check_output('apt-cache depends --recurse --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances --installed scylla', stderr=subprocess.STDOUT, shell=True, encoding='utf-8').splitlines()
+        pkgs=[]
+        for d in deps:
+            if re.match(r'^\w', d) and not re.match(r'.+:i386$', d) and d not in pkgs:
+                pkgs.append(d)
+        with open('{}/scylla-packages.txt'.format(homedir), 'w') as f:
+            for p in pkgs:
```
bentsi, line 183:
```suggestion
            for pkg_name in pkgs:
```

aws/ami/scylla.json
```diff
@@ -80,6 +80,12 @@
         "sudo {{user `python`}} /home/{{user `ssh_username`}}/scylla_install_ami {{ user `install_args` }}"
       ],
       "type": "shell"
+    },
+    {
+      "source": "/home/{{user `ssh_username`}}/scylla-packages.txt",
```
bentsi, line 85: why are you downloading the file to the build machine?

syuu1228: Please see description at https://github.com/scylladb/scylla-pkg/issues/1869, this intended to upload `scylla-packages.txt` to S3 (somewhere at downloads.scylladb.com), after finishing build of the AMI on Jenkins job.
The uploading part will be implement on Jenkins job after succeeded building the AMI, not this script.
So we need to do here is moving generated `scylla-packages.txt` from temporarily created AMI instance to the Jenkins node.

aws/ami/files/scylla_install_ami
syuu1228: Currently we generating recursive dependency list, not all installed packages list, so output of `rpm -qa` and `repoquery --requires --resolve --recursive scylla' are different.
However, I'm not really sure which one is correct, since our motivation is to align installed package between nodes, current way may cause package version difference out side of the dependency list.

@tzach any idea?

aws/ami/files/scylla_install_ami
syuu1228: On Ubuntu, there is no tool to easily generate recursive dependency list, but we almost succeeded to do that using `apt-cache depends`, and filter out the output.
However, the output contains not installed i386 version of packages even I specified option to `--installed`.
I verified all listed packages are installed, and seems like only i386 versions are not installed: 
```
$ for p in `apt-cache depends --recurse --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances --installed scylla|grep "^\w"`;do dpkg -l $p > /dev/null;done
dpkg-query: no packages found matching procps:i386
dpkg-query: no packages found matching util-linux:i386
dpkg-query: no packages found matching login:i386
dpkg-query: no packages found matching openssl:i386
dpkg-query: no packages found matching dpkg:i386
dpkg-query: no packages found matching coreutils:i386
dpkg-query: no packages found matching libpam-modules-bin:i386
dpkg-query: no packages found matching tar:i386
dpkg-query: no packages found matching install-info:i386
```

Probably amd64 version of them are in dependency list, and installed.
But we don't need i386 version.
I guess there is an apt parameter to stop showing up them, but I haven't find it yet.
So the script is workaround of this problem.

syuu1228: Ah, I found another problem on CentOS version.
CentOS version does not has problem we have on Ubuntu version, but we forgetting to specify --installed option to repoquery.
So CentOS version output is including some of not installed packages.

syuu1228: Pushed fix for CentOS version. (add --installed)

aws/ami/files/scylla_install_ami
syuu1228: Your suggestion merged into newly pushed version.

aws/ami/files/scylla_install_ami
syuu1228: Your suggestion merged into newly pushed version.

syuu1228:
@avikivity Can we merge this now? Or is there still need to change?

avikivity:
@syuu1228 I don't maintain this repository. I'm happy to review if needed, but someone else will need to merge it.

penberg:
I am fine with this, but would prefer @bentsi or @fruch to ACK before merging.

syuu1228:
@yaronkaikov Rebased with latest master.

Note that I keep CentOS code, and also this supports gce and azure, not just aws.
Because we define file provisioner in scylla.json to download scylla-packages.txt, it executes unconditionally.
If we want to skip doing this on CentOS (or non aws target), we probably need to create empty scylla-packages.txt on the target.

yaronkaikov (approved):
LGTM

PR #98, opened 2021-02-18
fix(scylla_create_devices): improve logic for found devices

aleksbykov:
scylla_create_devices support 3 variants for data_device:
- auto: for aws instance will search nvme(ephemeral) device,
          if not found will search ebs volumes(ebs).
        for gce instances will search nvme(attached) devices
for aws instances:
- attached: will search ebs volumes(attached) and if such volumes
            added to os as nvme devices, will use them
- instance_store: will search nvme(ephemeral) devices and use

### DIFF
--- a/common/scylla_configure.py
+++ b/common/scylla_configure.py
@@ -45,7 +45,16 @@ class ScyllaMachineImageConfigurator:
         'developer_mode': False,
         'post_configuration_script': '',
         'post_configuration_script_timeout': 600,  # seconds
-        'start_scylla_on_first_boot': True
+        'start_scylla_on_first_boot': True,
+        'data_device': 'auto'  # Supported options:
+                               #   instance_store - find all ephemeral devices (only for AWS)
+                               #   attached - find all attached devices and use them
+                               #   auto - automatically select devices using following strategy:
+                               #       GCE: select attached NVMe.
+                               #       AWS:
+                               #           if ephemeral found - use them
+                               #           else if attached EBS found use them
+                               #           else: fail create_devices
     }
 
     DISABLE_START_FILE_PATH = Path("/etc/scylla/ami_disabled")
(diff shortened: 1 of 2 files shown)


common/scylla_configure.py
```diff
@@ -146,11 +155,10 @@ def start_scylla_on_first_boot(self):
...
 
     def create_devices(self):
-        device_type = self.instance_user_data.get("data_device")
-        use_device_type = "--data-device {}".format(device_type) if device_type else ""
-
+        device_type = self.instance_user_data.get("data_device", self.CONF_DEFAULTS['data_device'])
         try:
-            subprocess.run("/opt/scylladb/scylla-machine-image/scylla_create_devices {}".format(use_device_type), shell=True, check=True)
+            LOGGER.info(f"Create scylla data devices as {device_type}")
+            subprocess.run("/opt/scylladb/scylla-machine-image/scylla_create_devices --data-device {}".format(device_type), shell=True, check=True)
```
bentsi, line 161: Use f-string:
```suggestion
            subprocess.run(f"/opt/scylladb/scylla-machine-image/scylla_create_devices --data-device {device_type}", shell=True, check=True)
```

common/scylla_create_devices
```diff
@@ -36,12 +36,14 @@ def create_raid(devices):
...
-            return [os.path.join('/dev/', name) for name in instance.ebs_disks()]
-        elif device_type == "instance_store":
-            return [os.path.join('/dev/', name) for name in instance.ephemeral_disks()]
-        else:
-            raise Exception(f"Unknown devicetype {device_type}")
+            devices = [str(Path('/dev', name)) for name in instance.ebs_disks() if Path('/dev', name).exists()]
+        if not devices or device_type == "instance_store":
+            devices = [str(Path('/dev', name)) for name in instance.ephemeral_disks()]
+        if not devices:
+            raise Exception(f"No devices attached or instance_store were found devicetype")
```
bentsi, line 45:
```suggestion
            raise Exception(f"No block devices were found for '{device_type}' device type")
```

bentsi (changes requested):
1) please address comments
2) did you check that `instance.ebs_disks()` and `instance.ephemeral_disks()` returns correct values?

common/scylla_create_devices
aleksbykov: Done

common/scylla_configure.py
aleksbykov: Done

aleksbykov:
> 1. please address comments

Done.
> 2. did you check that `instance.ebs_disks()` and `instance.ephemeral_disks()` returns correct values?

Yes, i had several instances with different configurations (ebs only as nvme, ebs only as attached, nvme+ebs) and i uploaded these files and check that disks correctly found and used. 
Also, i want to create new custom ami and check that scylla installed correctly

bentsi (approved):

bentsi:
can you create AMI with this code and verify that this works?

bentsi:
backported to 4.4

PR #101, opened 2021-03-09
gce: fix --localrpm build

syuu1228:
--localrpm mode is broken, due to calling wrong function name.
Fix to correct one.


syuu1228:
ping

bentsi (approved):

PR #104, opened 2021-03-15
aws/ami: fix broken Docker build for CentOS AMI

syuu1228:
build_with_docker.sh currently fails to call ../../SCYLLA-VERSION-GEN,
fixed the issue to make it runnable again.


syuu1228:
ping

bentsi (approved):

PR #113, opened 2021-03-25
dist/debian: fix renaming debian/scylla-* files rule

syuu1228:
Ported https://github.com/scylladb/scylla/commit/d7f202f9008ce312a8d0b71b41d073db7923ce1c to scylla-machine-image to fix scylladb/scylla-enterprise-machine-image#16.

----

Current renaming rule of debian/scylla-* files is buggy, it fails to
install some .service files when custom product name specified.

Introduce regex based rewriting instead of adhoc renaming, and fixed
wrong renaming rule.

Fixes scylladb/scylla-enterprise-machine-image#16

### DIFF
--- a/dist/debian/debian_files_gen.py
+++ b/dist/debian/debian_files_gen.py
@@ -23,8 +23,9 @@
 
 import string
 import os
-import glob
 import shutil
+import re
+from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


syuu1228:
ping

bentsi (approved):

PR #117, opened 2021-04-05
dist: remove .bash_profile from package, configure it in scylla_install_ami

syuu1228:
We have permission issue on /home/scyllaadm/, it owned by root:root but
should be scyllaadm:scyllaadm.
However, we create scyllaadm user AFTER installing scylla-machine-image
package, so it's not possible to do that in the package.

We don't really need to keep /home/scyllaadm/.bash_profile in package.
We just need to do "echo /opt/scylladb/scylla-machine-image/scylla_login >> ~/.bash_profile" in scylla_install_ami instead.
In fact, we already do so in GCE since login user of GCE is not static,
we addd scylla_login to /etc/skel/.bash_profile.
So let's align the way to configure .bash_profile to GCE, drop
.bash_profile from rpm/deb packages, configure .bash_profile in
scylla_install_ami.

Also, we need to take care upgrade scylla-machine-image package.
To do so, remove /home/scyllaadm/.bash_profile symlink and generate new
.bash_profile in %posttrans of the rpm package.

Fixes scylladb/scylla-enterprise#1684

### DIFF
--- a/aws/ami/files/scylla_install_ami
+++ b/aws/ami/files/scylla_install_ami
@@ -145,8 +145,13 @@ if __name__ == '__main__':
     run('cloud-init clean')
     run('cloud-init init')
     for skel in glob.glob('/etc/skel/.*'):
-        if not skel.endswith('.bash_profile'):
-            shutil.copy(skel, '/home/scyllaadm')
+        shutil.copy(skel, '/home/scyllaadm')
+    if distro == 'centos':
+        profile = '/home/scyllaadm/.bash_profile'
+    else:
+        profile = '/home/scyllaadm/.profile'
+    with open(profile, 'a') as f:
+        f.write('\n\n/opt/scylladb/scylla-machine-image/scylla_login\n')
     run('useradd -o -u 1000 -g scyllaadm -s /bin/bash -d /home/scyllaadm centos')
     if distro == 'ubuntu':
         run('useradd -o -u 1000 -g scyllaadm -s /bin/bash -d /home/scyllaadm ubuntu')
--- a/dist/redhat/scylla-machine-image.spec
+++ b/dist/redhat/scylla-machine-image.spec
@@ -41,9 +41,6 @@ install -m755 common/scylla_configure.py common/scylla_create_devices \
     --with-python3 ${RPM_BUILD_ROOT}/opt/scylladb/python3/bin/python3 \
     common/scylla_image_setup common/scylla_login common/scylla_configure.py \
     common/scylla_create_devices
-install -d -m755 $RPM_BUILD_ROOT/home
-install -d -m755 $RPM_BUILD_ROOT/home/scyllaadm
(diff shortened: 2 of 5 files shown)


dist/redhat/scylla-machine-image.spec
```diff
@@ -58,6 +55,16 @@ install -m755 common/.bash_profile $RPM_BUILD_ROOT/home/scyllaadm
 %postun
 %systemd_postun scylla-image-setup.service
 
+%posttrans
+if [ -L /home/scyllaadm/.bash_profile ] && [ ! -e /home/scyllaadm/.bash_profile ]; then
+    rm /home/scyllaadm/.bash_profile
+    cp /etc/skel/.bash_profile /home/scyllaadm/
+    chown scyllaadm:scyllaadm /home/scyllaadm/.bash_profile
+    echo "" >> /home/scyllaadm/.bash_profile
+    echo "" >> /home/scyllaadm/.bash_profile
```
bentsi, line 64: same line as above...

bentsi: and it does nothing

roydahan: @syuu1228 ping

syuu1228: It is just extra new lines for .bash_profile readability, can drop since it does noting

amoskong: > It is just extra new lines for .bash_profile readability, can drop since it does noting

echo -e '\n' >> /home/scyllaadm/.bash_profile

syuu1228: Changed to `echo -e '\n'`, suggested by @amoskong

syuu1228:
ping

aws/ami/files/scylla_install_ami
```diff
@@ -145,8 +145,13 @@ if __name__ == '__main__':
...
+        shutil.copy(skel, '/home/scyllaadm')
+    if distro == 'centos':
+        profile = '/home/scyllaadm/.bash_profile'
+    else:
+        profile = '/home/scyllaadm/.profile'
+    with open(profile, 'a') as f:
+        f.write('\n\n/opt/scylladb/scylla-machine-image/scylla_login\n')
     run('useradd -o -u 1000 -g scyllaadm -s /bin/bash -d /home/scyllaadm centos')
     if distro == 'ubuntu':
         run('useradd -o -u 1000 -g scyllaadm -s /bin/bash -d /home/scyllaadm ubuntu')
```
bentsi, line 157: for Ubuntu useradd will be executed twice...

syuu1228: It is existing code not added by this PR so we need to create another issue or PR to discuss, but anyway,
Maybe we don't need 'ubuntu' alias since we use 'scyllaadm' from initial version of the AMI?
We decided to add 'centos' alias to Ubuntu AMI to keep compatibility, so this one is needed.

bentsi (approved):

PR #129, opened 2021-04-26
Fail ami in script

hagitsegev:
The AMI side of #110 
This is done together with scylladb/scylla-pkg#2078
I want the AMI pipeline to get an error when AMI creation failed.
We also run the process with tee to log file, so we can read the ID from it.
This is a duplication, as the AMI script already write logs.

Get the name of the log from caller (still supply the current value as default). This will prevent the need to save a copy of the log (tee on tee).
Search the log for error, and return status.

Also align these scripts to the GCE scripts.

### DIFF
--- a/aws/ami/build_deb_ami.sh
+++ b/aws/ami/build_deb_ami.sh
@@ -18,23 +18,27 @@ source ../../SCYLLA-VERSION-GEN
 
 PRODUCT=$(cat build/SCYLLA-PRODUCT-FILE)
 DIR=$(dirname $(readlink -f $0))
+EXIT_STATUS=0
+DRY_RUN=false
 
 print_usage() {
     echo "build_deb_ami.sh --localdeb --repo [URL] --target [distribution]"
-    echo "  --localdeb  deploy locally built debs"
-    echo "  --repo  repository for both install and update, specify .repo/.list file URL"
+    echo "  --localdeb            deploy locally built debs"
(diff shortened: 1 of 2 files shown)


yaronkaikov (approved):

bentsi:
@hagitsegev why there are 2 commits?

aws/ami/build_deb_ami.sh
```diff
@@ -18,23 +18,27 @@ source ../../SCYLLA-VERSION-GEN
...
     echo "  --dry-run             validate template only (image is not built)"
     echo "  --download-no-server  download all deb needed excluding scylla from repo-for-install"
+    echo "  --log-file            Path for log. Default build/ami.log on current dir"
     exit 1
 }
 LOCALDEB=0
 DOWNLOAD_ONLY=0
-PACKER_SUB_CMD="build"
-
+PACKER_SUB_CMD="build -force -on-error=abort"
```
bentsi, line 38: why do you add "force" option?

hagitsegev: Hmmmm...
I did this also for GCE (also use packer).
Reading the docs https://www.packer.io/docs/commands/build
It looks good:

Forces a builder to run when artifacts from a previous build prevent a build from running. The exact behavior of a forced build is left to the builder. In general, a builder supporting the forced build will remove the artifacts from the previous build. This will allow the user to repeat a build without having to manually clean these artifacts beforehand.

PR #130, opened 2021-04-30
scylla_login: print warning message when no RAID disks available

syuu1228:
Scylla AMI requires additional disks to construct data volume, so our
AMI image automatically attaches ephemeral disks for it.
However, some unsupported additional instance types does not supported
any additional disks, so AMI will launch without any additional disks.
On such environment, we have added readable error message on 37ec32c,
but it still only able to check on journalctl.

To warn users more clearly, add new warning message on login prompt.

Fixes scylladb/scylla#8501


penberg (approved):
Looks good to me.

syuu1228:
ping

syuu1228:
ping

syuu1228:
ping

PR #135, opened 2021-05-24
gce-image: set uniqu image name

yaronkaikov:
got the following error during gce-image build:
```
03:09:46  --> googlecompute: Error creating instance: googleapi: Error 409: The resource 'projects/scylla-images/zones/europe-west1-b/instances/scylla-4-6-dev-0-20210521-a2379b96b1-1' already exists, alreadyExists
```

This is since we are building from same package version and the image
name is not unique.

Let's make sure each gce image buid has a unique name

Fixes: 	https://github.com/scylladb/scylla-pkg/issues/1879

### DIFF
--- a/gce/image/build_image.sh
+++ b/gce/image/build_image.sh
@@ -182,7 +182,6 @@ echo "SCYLLA_JMX_VERSION: $SCYLLA_JMX_VERSION"
 echo "SCYLLA_TOOLS_VERSION: $SCYLLA_TOOLS_VERSION"
 echo "SCYLLA_PYTHON3_VERSION: $SCYLLA_PYTHON3_VERSION"
 echo "BUILD_ID: $BUILD_ID"
-echo "WORKING_BRANCH: $BRANCH_VERSION"
 echo "Calling Packer..."
 
 /usr/bin/packer ${PACKER_SUB_CMD} \
@@ -196,7 +195,6 @@ echo "Calling Packer..."
   -var scylla_tools_version="$SCYLLA_TOOLS_VERSION" \
   -var scylla_python3_version="$SCYLLA_PYTHON3_VERSION" \
   -var scylla_build_id="$BUILD_ID" \
-  -var scylla_branch_version="$BRANCH_VERSION" \
   scylla_gce.json
 
 # For some errors packer gives a success status even if fails.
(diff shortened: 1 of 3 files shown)


gce/image/scylla_gce.json
```diff
@@ -11,7 +11,7 @@
       "machine_type": "{{user `instance_type`}}",
       "metadata": {"block-project-ssh-keys": "TRUE"},
       "image_family": "scylla",
-      "image_name": "scylla-{{user `scylla_branch_version`| clean_resource_name}}-build-{{user `scylla_build_id`| clean_resource_name}}",
+      "image_name": "scylla-{{user `scylla_image_name`}}-build-{{user `scylla_build_id`}}",
```
bentsi, line 14: why did you remove `clean_resource_name` it 

> clean_resource_name will convert upper cases to lower cases and replace illegal characters with a "-" character. 

https://www.packer.io/docs/templates/legacy_json_templates/engine#clean_resource_name

yaronkaikov: By mistake. fixed

gce/image/build_image.sh
```diff
@@ -163,6 +162,7 @@ else
 
 fi
 
+SCYLLA_IMAGE_NAME=$(echo $SCYLLA_VERSION | sed "s/\./-/g")
```
bentsi, line 165: are you sure you need the sed part? `clean_resource_name` will normalize the name..

bentsi: and better to put $(...) in quotes

yaronkaikov: Let me verify it

yaronkaikov: @bentsi you were right. it's not needed. removed

gce/image/build_image.sh
```diff
@@ -18,7 +18,6 @@ DRY_RUN=false
 source ../../SCYLLA-VERSION-GEN
 
 PRODUCT=$(cat build/SCYLLA-PRODUCT-FILE)
-BUILD_ID=$(date -u '+%FT%H-%M-%S')
```
bentsi, line 21: if you delete the build_id how it will be passed in line 195?

yaronkaikov: we provide it as a parameter

yaronkaikov: I can set maybe a default value like - `000` or something

bentsi: let's keep it

bentsi (approved):

PR #139, opened 2021-06-10
aws: create symlinks for /home/centos on Ubuntu AMI

syuu1228:
We need to create /home/centos symlink on Ubuntu AMI, since we have
centos alias user for compatibility.

Fixes scylladb/scylla-enterprise-machine-image#22


slivne (approved):
LGTM

PR #147, opened 2021-06-30
scylla.yaml:security Groups ingress: set to tcp

yaronkaikov:
Today we set ipProtocol to `-1` which allow inbound traffic of all
types. Let's specify it to `tcp`

based on 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html

Fixes: 
https://github.com/scylladb/scylla-enterprise-machine-image/issues/32


penberg:
@yaronkaikov Let's open issues against open source so people outside our company can see them. We can refer to the upstream issues from the enterprise ones.

penberg:
It would be good to understand why @fruch set `MapPublicIpOnLaunch` to `true` in the first place (the default is `false`).

penberg:
It's better to spell out what `MapPublicIpOnLaunch` does in the commit message title. For example, `cloudformation: Don't map public IP on launch`, and then in the description mention the `AWS::EC2::Subnet:MapPublicIpOnLaunch` parameter.

yaronkaikov:
> It's better to spell out what `MapPublicIpOnLaunch` does in the commit message title. For example, `cloudformation: Don't map public IP on launch`, and then in the description mention the `AWS::EC2::Subnet:MapPublicIpOnLaunch` parameter.

updated

yaronkaikov:
@fruch @penberg Looks like when i change MapPublicIpOnLaunch to false, the cluster is not comming up. Any idea why?

fruch:
> It would be good to understand why @fruch set `MapPublicIpOnLaunch` to `true` in the first place (the default is `false`).

no real reason, it was copy pasted from the cloud back then.

penberg:
@yaronkaikov Perhaps Scylla is _using_ that public IP address somewhere, and we should switch to the private one instead?

yaronkaikov:
@penberg Well it could be , but i couldn't find any mention for this public IP. so i suspect it's related to the actuall template . looks like the public IP is needed in order to perform the following which are part of the template: 
```
{%- if new_ami_boot %}
      UserData: !Base64
        'Fn::Join':
          - ''
          - - '{"scylla_yaml": {"seed_provider": [{"class_name": "org.apache.cassandra.locator.SimpleSeedProvider", "parameters": [{"seeds": "'
            - !Join
              - ','
              - - !If [Launch1, '172.31.0.11', !Ref "AWS::NoValue"]
                - !If [Launch2, '172.31.1.11', !Ref "AWS::NoValue"]
                - !If [Launch3, '172.31.2.11', !Ref "AWS::NoValue"]
            - '"}]}]'
            - !Sub ', "cluster_name": "${ClusterName}", '
            - '"endpoint_snitch": "org.apache.cassandra.locator.Ec2Snitch"}, '
            - '"start_scylla_on_first_boot": true, '
            - '"post_configuration_script" : "'
            - !Base64
              'Fn::Join':
                - ''
                - - !Sub |
                    #!/bin/bash -ex
                    sudo apt-get update
                    sudo apt-get -y install python3-pip
                    sudo pip install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz

                    /usr/local/bin/cfn-signal --exit-code 0 --resource Node{{ node_index }} --region ${AWS::Region} --stack ${AWS::StackName}
            - '"}'
{%- else %}
      UserData: !Base64
        'Fn::Join':
          - ''
          - - !Sub '--clustername ${ClusterName}'
            - |+

            - '--base64postscript '
            - !Base64
              'Fn::Join':
                - ''
                - - !Sub |
                    #!/bin/bash -ex

                    sudo apt-get update
                    sudo apt-get -y install python3-pip
                    sudo pip install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz

                    trap '/usr/local/bin/cfn-signal --exit-code 1 --resource Node{{ node_index }} --region ${AWS::Region} --stack ${AWS::StackName}' ERR

                    export PUBLIC_IP=${Node2ElasticIP}
                    export PRIVATE_IP=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
                    sed -i "s|# broadcast_address: 1.2.3.4|broadcast_address: $PRIVATE_IP|g" /etc/scylla/scylla.yaml
                  - |+

                  - export SEEDS=
                  - !Join
                    - ','
                    - - !If [Launch1, '172.31.0.11', !Ref "AWS::NoValue"]
                      - !If [Launch2, '172.31.1.11', !Ref "AWS::NoValue"]
                      - !If [Launch3, '172.31.2.11', !Ref "AWS::NoValue"]
                  - |+

                  - !Sub |
                    sed -i "s|          - seeds: \"$PRIVATE_IP\"|          - seeds: \"$SEEDS\"|g" /etc/scylla/scylla.yaml
                    /usr/local/bin/cfn-signal --exit-code 0 --resource Node{{ node_index }} --region ${AWS::Region} --stack ${AWS::StackName}
                  - |+

            - |+

            - |
              --totalnodes 1 --reflector example.com
{%- endif %}

```

penberg:
@yaronkaikov There's a typo in the subject line: s/lunch/launch/

penberg:
@yaronkaikov Yeah, looks like it uses those public IP addresses to find seed nodes (for forming a cluster). I am not sure how this is supposed to be fixed, though. @fruch, any ideas?

yaronkaikov:
> @yaronkaikov There's a typo in the subject line: s/lunch/launch/

Fixed

yaronkaikov:
> @yaronkaikov Yeah, looks like it uses those public IP addresses to find seed nodes (for forming a cluster). I am not sure how this is supposed to be fixed, though. @fruch, any ideas?

So i think i just figured it out. during the cloudformation creating we first installing
```
sudo pip install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz
```
And then running what should actually indicate that the node is up
```
/usr/local/bin/cfn-signal --exit-code 0 --resource Node{{ node_index }} --region ${AWS::Region} --stack ${AWS::StackName}
```
But , because we removed the public IP we can't install it anymore. So what i did is to move the installation part to AMI creation. 

Testing it now, but it make sense (at least to me)

yaronkaikov:
There are 2 options for solution here. 
1) Is to completely disable the public IP for all nodes. meaning that user can set a cluster but he will not be able to access it unless he already has a VPN service
2) Adding a parameter allowing the user to choose if he want to set a cluster with public IP access. default will be false. If he decided to set this parameter to **True** then we will set a public IP only for first cluster node. 

@tzach @mailmahee @slivne  Which solution it is?

mailmahee:
@yaronkaikov I think option 2 is better. The question I have is Why do you think only one public IP is needed? The client will only be able to establish CQL connection to only one node...

yaronkaikov:
> @yaronkaikov I think option 2 is better. The question I have is Why do you think only one public IP is needed? The client will only be able to establish CQL connection to only one node...

@mailmahee Before our change each node got a public IP address. You requested not to map it. So , if we enable Public ip for each node, there is nothing that need to be change. because this is what we configured from the first place

mailmahee:
I think the request is to keep the default to false and not have public IP mapped - and if the customer wants to set a public IP then he should be able to.

yaronkaikov:
> I think the request is to keep the default to false and not have public IP mapped - and if the customer wants to set a public IP then he should be able to.

So i guess it wasn't clear from the original issue https://github.com/scylladb/scylla-enterprise-machine-image/issues/31 
I will revert the none relevant changes

mailmahee:
> > I think the request is to keep the default to false and not have public IP mapped - and if the customer wants to set a public IP then he should be able to.
> 
> So i guess it wasn't clear from the original issue [scylladb/scylla-enterprise-machine-image#31](https://github.com/scylladb/scylla-enterprise-machine-image/issues/31)
> I will revert the none relevant changes

Thanks, @yaronkaikov!  and sorry for any non-clear language in the issue, I will try to communicate clearer next time.

yaronkaikov:
> > > I think the request is to keep the default to false and not have public IP mapped - and if the customer wants to set a public IP then he should be able to.
> > 
> > 
> > So i guess it wasn't clear from the original issue [scylladb/scylla-enterprise-machine-image#31](https://github.com/scylladb/scylla-enterprise-machine-image/issues/31)
> > I will revert the none relevant changes
> 
> Thanks, @yaronkaikov! and sorry for any non-clear language in the issue, I will try to communicate clearer next time.

@mailmahee Not a problem. i have removed the none relevant changes. 
@penberg @tzach @slivne So i have added a parameter asking the user if to enable public access or not. default is not to allow access using public IP.
@penberg @bentsi Please review

aws/cloudformation/scylla.yaml.j2
```diff
@@ -177,33 +191,29 @@ Resources:
               --totalnodes 1 --reflector example.com
 {%- endif %}
       NetworkInterfaces:
-        - NetworkInterfaceId: !Ref Node{{ node_index }}ETH0
+        - AssociatePublicIpAddress: !If [EnablePublicAccess, true, false]
```
benipeled, line 194: I think it can be shortened by using the parameter value directly `!Ref EnablePublicAccess` and remove the `EnablePublicAccess` condition on line 88

benipeled (approved):
LGTM

bentsi (approved):

PR #148, opened 2021-07-01
fix(cfn): move to use ubuntu base repo as master does

fruch:
Since the AMI move to Ubuntu, we need to install different the cfn tools
on boot,

IDEA: make it support both repos


yaronkaikov (approved):
LGTM 
@fruch We don't need to support both OS AFAIK. only Ubuntu based

Issue #137, opened 2021-06-09
Feature: Support AMI image based on arm architecture


yaronkaikov:
Recently we have added support in Scylla to build relocatable and packages on machines based on ARM. Let's add also support for machine image so we can use an arm based instances

syuu1228:
We have merged aarch64 AMI patch https://github.com/scylladb/scylla-machine-image/pull/89
And also rpm & deb patch https://github.com/scylladb/scylla/pull/7026

I just tried to build aarch64 AMI (Ubuntu) with latest master, it seems still working (but with unsupported instance mode).

test image is -> us-east-1: ami-0ef4b45acf55d8f82
```
Welcome to Ubuntu 20.04.2 LTS (GNU/Linux 5.4.0-1035-aws aarch64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Mon Jun 14 10:57:17 UTC 2021

  System load:           1.46
  Usage of /:            55.7% of 28.91GB
  Memory usage:          0%
  Swap usage:            0%
  Processes:             573
  Users logged in:       0
  IPv4 address for eth0: 10.0.0.4
  IPv6 address for eth0: 2600:1f18:423:1700:58f8:657e:e8ea:cad1


11 updates can be applied immediately.
7 of these updates are standard security updates.
To see these additional updates run: apt list --upgradable



The programs included with the Ubuntu system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.


   _____            _ _       _____  ____  
  / ____|          | | |     |  __ \|  _ \ 
 | (___   ___ _   _| | | __ _| |  | | |_) |
  \___ \ / __| | | | | |/ _` | |  | |  _ < 
  ____) | (__| |_| | | | (_| | |__| | |_) |
 |_____/ \___|\__, |_|_|\__,_|_____/|____/ 
               __/ |                       
              |___/                        

Version:
       4.6.dev-0.20210614.061e43e9d
Nodetool:
	nodetool help
CQL Shell:
	cqlsh
More documentation available at: 
	http://www.scylladb.com/doc/
By default, Scylla sends certain information about this node to a data collection server. For information, see http://www.scylladb.com/privacy/

    c6gd is not eligible for optimized automatic tuning!
[...]

PR #153, opened 2021-07-16
AMI: Add support for ARM based AMI build

yaronkaikov:
Today we have an AMI based on x86 arch only. Now that we have ARM based
packaging , we want to have also AMI based on ARM

Fixes: #137

### DIFF
--- a/aws/ami/build_deb_ami.sh
+++ b/aws/ami/build_deb_ami.sh
@@ -117,9 +117,15 @@ check_deb_exists () {
     done
 }
 declare -A AMI
-AMI=(["x86_64"]=ami-0074ee617a234808d ["aarch64"]=ami-0ae74ae9c43584639)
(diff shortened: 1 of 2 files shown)


yaronkaikov:
@syuu1228 I need you assistance here. i managed to build the AMI image. but when loging i get the following error
```
Traceback (most recent call last):
  File "/opt/scylladb/scylla-machine-image/libexec/scylla_login", line 116, in <module>
    colorprint(MSG_UNSUPPORTED_INSTANCE_TYPE_NO_DISKS % cloud_instance.GETTING_STARTED_URL,
TypeError: not all arguments converted during string formatting
```
Can you have a look ? i have an instance id `i-0f49222dd39d2e8e0` (region us-east-1) and it's based on `i-0f49222dd39d2e8e0`

bentsi (approved):

Issue #157, opened 2021-07-16
 Cloudformation script parameter - VPC : assign the CIDR range for the new VPC


yaronkaikov:
AWS::EC2::VPC The user should be able to asign the CIDR range for the new VPC, Create a parameter asking this from the user

Severity: Critical

PR #162, opened 2021-07-16
Cloudformation VPC:assign CIDR range for new VPC

yaronkaikov:
Need to allow a user to be able to asign the CIDR range for the new VPC,
Let's Create a parameter asking this from the user

Fixes: https://github.com/scylladb/scylla-machine-image/issues/157

### DIFF
--- a/aws/cloudformation/scylla.yaml
+++ b/aws/cloudformation/scylla.yaml
@@ -56,6 +56,11 @@ Parameters:
     Type: 'AWS::EC2::KeyPair::KeyName'
     ConstraintDescription: must be the name of an existing EC2 KeyPair.
 
+  VpcCidrIp:
+    Description: Set CIDP IP range for VPC
+    Type: String
+    ConstraintDescription: must be valid subnet
+
 # Those conditions would be used to enable nodes based on InstanceCount parameter
 Conditions:
     Launch1: !Equals [1, 1]
@@ -866,7 +871,7 @@ Resources:
         - Key: ClusterName
           Value: !Ref ClusterName
       SecurityGroupIngress:
-        - CidrIp: 172.31.0.0/16
+        - CidrIp: !Ref VpcCidrIp
           FromPort: 22
           ToPort: 22
           IpProtocol: '-1'
@@ -887,7 +892,7 @@ Resources:
     Type: 'AWS::EC2::Subnet'
     Properties:
       AvailabilityZone: !Ref AvailabilityZone
-      CidrBlock: 172.31.0.0/16
+      CidrBlock: !Ref VpcCidrIp
       MapPublicIpOnLaunch: true
       Tags:
         - Key: Name
@@ -905,7 +910,7 @@ Resources:
   VPC:
     Type: 'AWS::EC2::VPC'
     Properties:
-      CidrBlock: 172.31.0.0/16
+      CidrBlock: !Ref VpcCidrIp
       Tags:
         - Key: Name
           Value: !Sub '${ClusterName}-VPC'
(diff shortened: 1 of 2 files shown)


aws/cloudformation/scylla.yaml.j2
```diff
@@ -61,6 +61,11 @@ Parameters:
     Type: 'AWS::EC2::KeyPair::KeyName'
     ConstraintDescription: must be the name of an existing EC2 KeyPair.
 
+  VpcCidrIp:
+    Description: Set CIDP IP range for VPC
+    Type: String
+    ConstraintDescription: must be vaild subnet
```
penberg, line 67: s/vaild/valid/

aws/cloudformation/scylla.yaml
```diff
@@ -56,6 +56,11 @@ Parameters:
     Type: 'AWS::EC2::KeyPair::KeyName'
     ConstraintDescription: must be the name of an existing EC2 KeyPair.
 
+  VpcCidrIp:
+    Description: Set CIDP IP range for VPC
+    Type: String
+    ConstraintDescription: must be vaild subnet
```
penberg, line 62: s/vaild/valid/

penberg: Is this sufficient, though? We have mentions of `172.31.x.x` IP addresses elsewhere in the CloudFormation template. Surely those need to be parametrised too for this to make sense?

/cc @fruch

yaronkaikov: @penberg looks like we do need to parametrised those as well. but what if user will create a 5 node cluster?

yaronkaikov: @fruch @penberg ping

aws/cloudformation/scylla.yaml.j2
```diff
@@ -61,6 +61,11 @@ Parameters:
     Type: 'AWS::EC2::KeyPair::KeyName'
     ConstraintDescription: must be the name of an existing EC2 KeyPair.
 
+  VpcCidrIp:
+    Description: Set CIDP IP range for VPC
+    Type: String
+    ConstraintDescription: must be valid subnet
```
fruch, line 67: how about doing `172.31.0.0/16` as default ?

aws/cloudformation/scylla.yaml
fruch: I'm not sure what the question here...

aws/cloudformation/scylla.yaml.j2
yaronkaikov: The requirement (for AWS marketplace i assume) is to configure the VPC range. User must supply it, i am not sure about default , @mailmahee , any input?

aws/cloudformation/scylla.yaml
yaronkaikov: in the template file we have those
```
          - - '{"scylla_yaml": {"seed_provider": [{"class_name": "org.apache.cassandra.locator.SimpleSeedProvider", "parameters": [{"seeds": "'
            - !Join
              - ','
              - - !If [Launch1, '172.31.0.11', !Ref "AWS::NoValue"]
                - !If [Launch2, '172.31.1.11', !Ref "AWS::NoValue"]
                - !If [Launch3, '172.31.2.11', !Ref "AWS::NoValue"]
 ```
So if a user is using different subnet, we should probably also modify those values  ,right ?

yaronkaikov:
@tzach Now that we want to give a user the option to configure the VPC CIDR , we have a new issue related to the nodes private IP address.
Until now it was a fix address based on node_index, so it was easy to calculate. But now we don't know the subnet . 
We have an option is to add additional parameters to cloudformation asking for each node Private ip address (from the defined subnet).

What do you think? any better idea?

yaronkaikov:
This issue will be solved in https://github.com/scylladb/scylla-machine-image/pull/178

Issue #161, opened 2021-07-16
Cloudformation script parameter - SGCluster : Missing egress rule


yaronkaikov:
Missing egress rule in security group means CloudFormation will automatically add an outbound rule to allow all traffic outbound. Make this explicit by adding egress rule if it is the desired configuration. No need to parametize it.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#aws-properties-ec2-security-group--examples--Remove_Default_Rule

Severity: Recommended

PR #164, opened 2021-07-16
cloudformation: SGCluster : Missing egress rule

yaronkaikov:
Missing egress rule in security group means CloudFormation will 
automatically add an outbound rule to allow all traffic outbound. Make 
this explicit by adding egress rule if it is the desired configuration. 
No need to parametize it.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#aws-properties-ec2-security-group--examples--Remove_Default_Rule

Fixes: https://github.com/scylladb/scylla-machine-image/issues/161

### DIFF
--- a/aws/cloudformation/scylla.yaml
+++ b/aws/cloudformation/scylla.yaml
@@ -867,9 +867,7 @@ Resources:
           Value: !Ref ClusterName
       SecurityGroupIngress:
         - CidrIp: 172.31.0.0/16
-          FromPort: 22
-          ToPort: 22
-          IpProtocol: 'tcp'
+          IpProtocol: '-1'
       SecurityGroupEgress:
         - CidrIp: 0.0.0.0/0
           IpProtocol: '-1'
--- a/aws/cloudformation/scylla.yaml
+++ b/aws/cloudformation/scylla.yaml
@@ -870,6 +870,9 @@ Resources:
           FromPort: 22
           ToPort: 22
           IpProtocol: 'tcp'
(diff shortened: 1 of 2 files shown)


yaronkaikov:
@penberg ping

penberg:
If I understand the change correctly, this prevents a Scylla node from communicating with any external nodes. This means that the node will not be able to connect to other nodes in the cluster, no?

yaronkaikov:
> If I understand the change correctly, this prevents a Scylla node from communicating with any external nodes. This means that the node will not be able to connect to other nodes in the cluster, no?

@penberg Fixed. I have created a 3 node cluster using this cloudformation. i have 3 nodes up and running , but shouldn't i see under `nodetool status` all 3 of them?
Now i see
```
$ nodetool status

Datacenter: us-east
===================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address      Load       Tokens       Owns    Host ID                               Rack
UN  172.31.0.11  294.2 KB   256          ?       6b440f3e-a860-48b6-9545-d5d728c58920  1a
```

yaronkaikov:
@penberg So this is related to my change., i will figure how to fix it

yaronkaikov:
@penberg  please review again , apparently there was a configuration issue in https://github.com/scylladb/scylla-machine-image/pull/163 which actually messed up what i was trying to do here. Added another commit fixing it and explaining

PR #165, opened 2021-07-17
scylla_login: fix message to apply format parameter

syuu1228:
MSG_UNSUPPORTED_INSTANCE_TYPE_NO_DISKS is not applied format parameter,
even it passed with '%'.


syuu1228:
See https://github.com/scylladb/scylla-machine-image/pull/153#issuecomment-881259364

yaronkaikov:
@syuu1228 Thanks., confirm also it's working

yaronkaikov (approved):

PR #166, opened 2021-07-19
aws: on Ubuntu AMI, remove auto update features by default

### LINKED ISSUE #152: Unattended upgrades enabled by default
When additional software is installed on Ubuntu-based Scylla machine images, one may hit the following problem:

```
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 15976 (unattended-upgr)
```

Are unattended upgrades enabled by default due to some reason? What is the recommended way to disable it?

syuu1228:
We don't want to enable auto update features on our ami, remove
unattended-upgrades and update-notifier-common to disable them.

fixes #152

### DIFF
--- a/aws/ami/files/scylla_install_ami
+++ b/aws/ami/files/scylla_install_ami
@@ -110,6 +110,9 @@ if __name__ == '__main__':
         if args.repo_for_update:
             run('curl -L -o /etc/apt/sources.list.d/scylla.list {REPO_FOR_UPDATE}'.format(REPO_FOR_UPDATE=args.repo_for_update))
 
+        # disable unattended-upgrades
+        run('apt-get purge -y unattended-upgrades update-notifier-common')
+
     # use easy_install instead of pip, because easy_install requires
     # fewer dependencies, no need to install gcc, it makes our AMI smaller.
     if distro == 'centos':


slivne:
@adambabik / @nuivall / @carlosms this is the patch - still under review

adambabik (dismissed):
Looks good but it's hard to tell if it will work as expected. How was it tested?

Our workaround for now is disabling the `unattended-upgrades` service.

penberg (approved):
Looks good to me. @syuu1228 can you comment on how this was tested?

syuu1228:
> Looks good but it's hard to tell if it will work as expected. How was it tested?
> 
> Our workaround for now is disabling the `unattended-upgrades` service.

Correct, I realized that even after we set the APT config value, unattended-upgrades service is still running and seems like do something.
I think better solution is completely remove these auto update features.
So I pushed new version, that removes `unattended-upgrades update-notifier-common` packages.

After removed them, no unattended-upgrades service running.
It sill run apt-daily.timer / apt-daily-upgrade.timer but the service do nothing because APT::Periodic::Enable set to null since we removed all config files for APT::Periodic.
(the behavior defined at line 532 of /usr/lib/apt/apt.systemd.daily, tested with `sh -x`)

PR #167, opened 2021-07-19
build_deb_ami.sh: Fix condition typo casuing ami build to fail

yaronkaikov:
In https://github.com/scylladb/scylla-machine-image/pull/153/commits/27223c6e7eb5ffb93fb5f535f299bafd17174ac3 we started to support AMI build based on ARM.

I missed a condition typo which broke the x86 based AMI build

Fixing it here

### DIFF
--- a/aws/ami/build_deb_ami.sh
+++ b/aws/ami/build_deb_ami.sh
@@ -121,7 +121,7 @@ AMI=(["x86_64"]=ami-0074ee617a234808d ["aarch64"]=ami-0763a1094de643002)
 REGION=us-east-1
 SSH_USERNAME=ubuntu
 
-if [ uname -m | grep x86_64]; then
+if uname -m | grep x86_64 ; then
   INSTANCE_TYPE="c4.xlarge"
 else
   INSTANCE_TYPE="a1.xlarge"


yaronkaikov:
Testing:
https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/626/ (x86 based AMI) - **OK**
https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/629/ (arm based AMI) - **OK**

penberg (approved):

aws/ami/build_deb_ami.sh
```diff
@@ -121,7 +121,7 @@ AMI=(["x86_64"]=ami-0074ee617a234808d ["aarch64"]=ami-0763a1094de643002)
 REGION=us-east-1
 SSH_USERNAME=ubuntu
 
-if [ uname -m | grep x86_64]; then
+if uname -m | grep x86_64 ; then
   INSTANCE_TYPE="c4.xlarge"
 else
```
penberg, line 126: As a follow up, we should explicitly check for aarch64 here, and make the `else` branch print out an "unsupported architecture" error message to future proof it.

yaronkaikov: Ok - will add it

PR #168, opened 2021-07-20
build_deb_ami: improve arch verification

yaronkaikov:
This is a follow-up fix to https://github.com/scylladb/scylla-machine-image/pull/167.

Let's explicitly check for aarch64 amd x86 and inform incase we have unsupported architecture

### DIFF
--- a/aws/ami/build_deb_ami.sh
+++ b/aws/ami/build_deb_ami.sh
@@ -121,11 +121,18 @@ AMI=(["x86_64"]=ami-0074ee617a234808d ["aarch64"]=ami-0763a1094de643002)
 REGION=us-east-1
 SSH_USERNAME=ubuntu
 
-if uname -m | grep x86_64 ; then
-  INSTANCE_TYPE="c4.xlarge"
-else
-  INSTANCE_TYPE="a1.xlarge"
-fi
+arch="$(uname -m)"
+case "$arch" in
+  "x86_64")
+    INSTANCE_TYPE="c4.xlarge"
+    ;;
+  "aarch64")
+    INSTANCE_TYPE="a1.xlarge"
+    ;;
+  *)
+    echo "Unsupported architecture: $arch"
+    exit 1
+esac
 
 if [ $LOCALDEB -eq 1 ]; then
     INSTALL_ARGS="$INSTALL_ARGS --localrpm"


aws/ami/build_deb_ami.sh
```diff
@@ -123,8 +123,11 @@ SSH_USERNAME=ubuntu
 
 if uname -m | grep x86_64 ; then
   INSTANCE_TYPE="c4.xlarge"
-else
+elif uname -m | grep aarch64 ; then
   INSTANCE_TYPE="a1.xlarge"
+else
+  echo "Unsupported architecture"
```
penberg, line 129: We should print out the architecture here.

You could simplify the whole thing to:

You could simplify this to:

```bash
arch="$(uname -m)"
case "$arch" in
  "x86_64")
    INSTANCE_TYPE="c4.xlarge"
    ;;
  "aarch64")
    INSTANCE_TYPE="a1.xlarge"
    ;;
  *)
    echo "Unsupported architecture: $arch"
    exit 1
esac
```

penberg (changes requested):

aws/ami/build_deb_ami.sh
```diff
@@ -123,8 +123,11 @@ SSH_USERNAME=ubuntu
```
penberg, line 123: Commit subject line as an extra "a" at the beginning.

yaronkaikov:
AMI based on x86: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/630/ - **OK**
AMI based on arm: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/632/ - **OK**

aws/ami/build_deb_ami.sh
yaronkaikov: removed

aws/ami/build_deb_ami.sh
```diff
@@ -121,11 +121,18 @@ AMI=(["x86_64"]=ami-0074ee617a234808d ["aarch64"]=ami-0763a1094de643002)
...
+  "x86_64")
+    INSTANCE_TYPE="c4.xlarge"
+    ;;
+  "aarch64")
+    INSTANCE_TYPE="a1.xlarge"
+    ;;
+  *)
+    echo "Unsupported architecture: $arch"
+    exit 1
+esac
```
penberg, line 135: Beautiful! 🤩

PR #132, opened 2021-05-10
Merge scylla_install_(ami|image|azure) to single script

syuu1228:
Deduplicate almost similar script for each cloud, and make it easier to
maintain.

PR #169, opened 2021-07-25
GCE: Add Operating system and branch labels

hagitsegev:
Fixes #2270 (Enhancement).
To be merged with: https://github.com/scylladb/scylla-pkg/pull/2276
Add "branch" label for GCE, as well operating-system like tags on AMI.


hagitsegev:
Testing:
https://jenkins.scylladb.com/view/nexts/job/scylla-master/job/releng-testing/job/gce-image/384/console looks OK on logs, but I also want to see the real label.

gce/image/scylla_gce.json
```diff
@@ -19,6 +19,8 @@
       "omit_external_ip": false,
       "disk_size": 20,
       "image_labels":  {
+          "branch": "{{`user branch`| clean_resource_name}}",
```
penberg, line 22: This should probably be named `scylla_branch`.

hagitsegev: It is making sense, but no.
Yaron set this on AMI, to differentiate between the "production" AMIs and our testing AMIs, so we have "master" or "branch-4.5" for the production builds and on the other hand pkg-branch such as "fix-gce-label" as an ID when we debug.
I agree that the name is misleading.
What a better name should be (both here, and for AMI tag, depends on QA - as need to change their query I guess)?

hagitsegev:
I convert to draft, as something is not working yet. Need to understand.

gce/image/scylla_gce.json
yaronkaikov: Since the QA are looking for scylla_branch, when ever we did debug sessions it was always pointing to scylla's stable branch. causing them to use our debug AMI instead of the offical one. 

So it's actually scylla-pkg branch

yaronkaikov:
@hagitsegev Let's wait first to https://github.com/scylladb/scylla-machine-image/pull/132 , then rebase and make sure the changes are alligned

gce/image/scylla_gce.json
penberg: What is the purpose of this information? Showing scylla.pkg git branch name does not seem like something we want our end user to see...

hagitsegev: These "internal" testing items should not be visible to end users. Are they? IF so - we should put these somewhere else.
I also think that "master" items should not be visible to end users... before we run any test on them. We can set them as visible after the test passes (as we have for other install ways - the "latest nightly unstable".

enaydanov:
@hagitsegev, can we add BUILD_ID to the labels too (with name `build-id`)? To have it same as for AMIs.

yaronkaikov:
@hagitsegev https://github.com/scylladb/scylla-machine-image/pull/132 was merged.

gce/image/scylla_gce.json
```diff
@@ -18,6 +18,8 @@
       "omit_external_ip": false,
       "disk_size": 20,
       "image_labels":  {
+          "branch": "{{user `branch`| clean_resource_name}}",
+          "operating_system": "{{user `operating_system`| clean_resource_name}}",
```
enaydanov, line 22: @hagitsegev, can we add `BUILD_ID` to the labels too (with name `build-id`)? To have it same as for AMIs.

```suggestion
          "operating_system": "{{user `operating_system`| clean_resource_name}}",
          "build-id": "{{user `scylla_build_id`| clean_resource_name}}",          
```

bentsi (changes requested):
@hagitsegev can you add `build-id` too?

gce/image/scylla_gce.json
hagitsegev: Yes.
The whole code was changed, as scripts were "merged" on https://github.com/scylladb/scylla-machine-image/commit/a7eac3d48bd3791d6f4cb6b24648e287248bbfb1
So I'll do this for all.

enaydanov: @hagitsegev thanks, but it looks like `build-id` is missed in the `image_labels` dict. Can you check?

yaronkaikov (approved):

yaronkaikov (dismissed):

bentsi (changes requested):
- Please address comment:
image
- resolve conflicts

gce/image/scylla_gce.json
hagitsegev: @enaydanov On which build did you check it?

packer/scylla.json
```diff
@@ -93,7 +93,10 @@
           "scylla_jmx_version": "{{user `scylla_jmx_version`| clean_resource_name}}",
           "scylla_tools_version": "{{user `scylla_tools_version`| clean_resource_name}}",
           "scylla_python3_version": "{{user `scylla_python3_version`| clean_resource_name}}",
-          "user_data_format_version": "2"
+          "build_id": "{{user `build_id`| clean_resource_name}}",
```
enaydanov, line 96: Please, set it to `build-id`, i.e., with dash and not underscore.

enaydanov (changes requested):

packer/scylla.json
enaydanov: My point is to have it in the same format as for AMI.

hagitsegev:
On AMI all tags are with dashes.
On GCE - with underscores.
I think we should set all labels of GCE to be with dashes too. Any idea if some process is using the existing labels?
It is ugly to have some with underscores, as existing, and some new with dashes.
@bentsi / @yaronkaikov

hagitsegev:
Discussed with @yaronkaikov 
Decided to add the new labels as build-id and operating-system like AMI as @enaydanov requested.
I opened few issues on SCT: https://github.com/scylladb/scylla-cluster-tests/issues/4023 and https://github.com/scylladb/scylla-cluster-tests/issues/4022 and here #230 and on pkg https://github.com/scylladb/scylla-pkg/issues/2440 to align the names.

hagitsegev:
image

bentsi (approved):

PR #178, opened 2021-08-08
cloudformation: allow custom VPC-CIDR for scylla cluster

### LINKED ISSUE #157: Cloudformation script parameter - VPC : assign the CIDR range for the new VPC

benipeled:
Following a (customer) request, the VPC CIDR should be adjustable,
Due to CFN syntax limitation (no math-operator) the node IPs are
hardcoded (10, 11, etc.) which means CIDR netmask must be one of 8,16,24

Fixes #157

### DIFF
--- a/aws/cloudformation/scylla.yaml
+++ b/aws/cloudformation/scylla.yaml
@@ -69,6 +69,23 @@ Parameters:
     Type: 'AWS::EC2::KeyPair::KeyName'
     ConstraintDescription: must be the name of an existing EC2 KeyPair.
 
+  CIDR:
+    Description: |
+        CIDR range for Scylla VPC (ex. 172.31.0.0/16), Currently supports 8, 16, or 24 netmask,
+        The node IPs will be x.x.x.10, x.x.x.11, x.x.x.12 etc.
+    Type: 'String'
+    Default: '172.31.0.0/16'
+    AllowedPattern: (\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})/(\d{1,2})
+    ConstraintDescription: must be a valid CIDR (ex. 172.31.0.0/16)
+
+  ScyllaSeedIPs:
+    Description: |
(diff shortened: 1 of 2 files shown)


benipeled:
**The CFN form:**

image

**Tested with following params:**

image

aws/cloudformation/scylla.yaml
```diff
@@ -56,6 +56,23 @@ Parameters:
     Type: 'AWS::EC2::KeyPair::KeyName'
     ConstraintDescription: must be the name of an existing EC2 KeyPair.
 
+  CIDR:
+    Description: |
+        CIDR range will be used by Scylla VPC (ex. 172.31.0.0/16),
+        the node IPs will be x.x.x.10, x.x.x.11, x.x.x.12 etc.
```
yaronkaikov, line 62: You assume here that the range that will be provied by the user has this ip addresses (the ending). Why not changing it so we will always use first ip in range for Node0 and so on?

yaronkaikov (changes requested):
Looks good, a general comment. Please add a fix tag for the open source issue we have.

aws/cloudformation/scylla.yaml
benipeled: Well, you right, - the IPs (10,11,12,13 etc.) are hard coded due to CFN limitations
\-  even if I'll have the last octet (ex. `!Select [3, !Split ['.', !Select [0, !Split ['/', !Ref CIDR]]]]`), CFN has no math-operators we can use to increase the result by x,

The math-operators and other custom function can be done by `custom-resource` or `macro` which require pre-configured lambda function which is different story.

yaronkaikov (changes requested):
Looks good, a general comment. Please add a fix tag for the open source issue we have. Let's also update the PR description to explain the logic

benipeled:
@yaronkaikov see the updated CIDR description

yaronkaikov (approved):
LGTM

PR #179, opened 2021-08-11
cloudformation: public ip mapping code cleanup

yaronkaikov:
In https://github.com/scylladb/scylla-machine-image/pull/147 i have  added a parameter for enabling public ip mapping.
Part of the added code, seems to be not relevant 
Removing it


benipeled (approved):
Let's cut the user costs :)
LGTM

Issue #180, opened 2021-08-11
Cloudformation: re-order and group the cfn form parameters


benipeled:
Should be done with `AWS::CloudFormation::Interface`

PR #183, opened 2021-08-11
Cloudformation: reorganize and clarify the form

benipeled:
This commit includes:
 - Re-order parameters by using AWS::CloudFormation::Interface
 - Add friendly name for parameters
 - A few description clarifications
 - Change the InstanceCount to be a string with list of 1-10

Fixes: #180 

@yaronkaikov please review.

**The form new design:**

image

### DIFF
--- a/aws/cloudformation/scylla.yaml.j2
+++ b/aws/cloudformation/scylla.yaml.j2
@@ -10,37 +10,76 @@ Description: >-
 
   Caution: password authentication isn't enabled by default
 
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: Scylla Parameters
+        Parameters:
+          - ScyllaClusterName
+          - ScyllaAmi
+          - ScyllaSeedIPs
+      - Label:
+          default: AWS Parameters
+        Parameters:
+          - InstanceType
+          - InstanceCount
+          - AvailabilityZone
+          - CIDR
+          - EnablePublicAccess
+          - PublicAccessCIDR
+          - KeyName
+    ParameterLabels:
+      ScyllaClusterName:
+        default: Scylla cluster name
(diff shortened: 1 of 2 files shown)


aws/cloudformation/scylla.yaml.j2
```diff
@@ -10,37 +10,76 @@ Description: >-
...
+          - AvailabilityZone
+          - CIDR
+          - EnablePublicAccess
+          - SSHLocation
+          - KeyName
+    ParameterLabels:
+      ScyllaClusterName:
+        default: Scylla cluster name
+      ScyllaAmi:
+        default: Scylla AMI for the EC2 instances
```
yaronkaikov, line 36: AMI already indicates it's Ec2 instances. Maybe change it to Scylla AMI id?

aws/cloudformation/scylla.yaml.j2
```diff
@@ -10,37 +10,76 @@ Description: >-
...
+      SSHLocation:
+        default: Allowed subnet for public access (SSH)
+
 Parameters:
-  ClusterName:
+  ScyllaClusterName:
     Type: String
 
-  PublicCidrAccess:
+  SSHLocation:
```
yaronkaikov, line 56: Why Location?

yaronkaikov (changes requested):

yaronkaikov:
@benipeled Just a general comment, if you want me to review, you can just asign me as a reviewer (right side). if you can't let me know so i will fix it for you

aws/cloudformation/scylla.yaml.j2
benipeled: No good answer + it shouldn't be service specific (SSH in this case),
Changed to `PublicAccessCIDR`

yaronkaikov (approved):

Issue #159, opened 2021-07-16
Cloudformation script parameter - Mappings : Add an entry in the map for every region where your product should be available


yaronkaikov:
Add an entry in the map for every region where your product should be available. The AMI should be the same as the one for us-east-1

Severity: Critical

yaronkaikov:
@mailmahee @ealonso63  can you explain this please? according to the documentation in https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html , we can't set a region to a specific AMI id since it is something the user need to provide.

Or maybe i miss understood what is need to be done here. please clarify

PR #185, opened 2021-08-15
cloudformation: add ami map per available region

yaronkaikov:
Today setting a new stack reuqired the user to enter the relevant AMI id depend on the region

Let's add an entry in the map for every region where your product should be available

Fixes: https://github.com/scylladb/scylla-machine-image/issues/159

### DIFF
--- a/aws/cloudformation/scylla.yaml.j2
+++ b/aws/cloudformation/scylla.yaml.j2
@@ -75,10 +75,6 @@ Parameters:
       - 'true'
     Default: 'false'
 
-  ScyllaAmi:
-    Type: 'AWS::EC2::Image::Id'
-    ConstraintDescription: Enter a valid Scylla AMI ID for your selected region
-
   InstanceType:
     Type: String
     Default: i3.large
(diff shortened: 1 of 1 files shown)


benipeled:
LGTM,
You probably need another PR for the `placeholder` replacement as part of the AMI build - please link it to this PR.

penberg:
@yaronkaikov Why is the `scylla.yaml` file removed in this pull request? How are the placeholders filled?

Let's have @bentsi/@fruch look at this too.

yaronkaikov:
> @yaronkaikov Why is the `scylla.yaml` file removed in this pull request? How are the placeholders filled?
> 
> Let's have @bentsi/@fruch look at this too.

We did some change in the process, until now scylla.yaml was constnat without any changes. and the upload was during the packages repo creation. Now since we have those placeholders for AMI ids. we need do it differently
So from now, during AMI creation we also copy the AMI to testing regions. during the copy we also update the AMI id's per region. Second place is during promotion , we will update the scylla.yaml.j2 template with currect values and only then generate `scylla.yaml` and upload to S3 (Same location as before)

penberg (approved):
lgtm, hopefully we'll also get @bentsi's blessing

slivne:
Yes I know we are changing and its a reuirment for AWS marketplaces - so its needed.

We will also provide a template that will not be bound and created for every version

yaronkaikov:
@slivne @bentsi i restoreed `scylla.yaml` without the placeholders. This will will be upload to same location as today regardless to the release (name of the file will be `scylla.yaml` without any version number).

This work will be done as part of https://github.com/scylladb/scylla-pkg/pull/2314

penberg (approved):

PR #186, opened 2021-08-16
gce: use unique instance_name

### LINKED ISSUE #184: GCE: GCE build often break with "Error creating instance: googleapi: Error 409: The resource 'projects/scylla-images/zones/xxxx/instances/scylla-xxxx' already exists''
I'm getting `Error creating instance: googleapi: Error 409: The resource 'projects/scylla-images/zones/xxxx/instances/scylla-xxxx' already exists` repeatedly.

Probably one reason is we run `packer build -force -on-error=abort`:
`-on-error=abort` means packer build instance will stay alive when build failed.
`-force` should remove if the instance already exists, but maybe it fails to remove.
https://github.com/scylladb/scylla-machine-image/blob/next/gce/image/build_deb_image.sh#L39

Another problem is probably `instance_name` on json.
I think we should use unique name, using `{{uuid}}`, current name can be conflict when we build image with same scylla revision.
On packer document it says:
_instance_name (string) - A name to give the launched instance. Beware that this must be unique. Defaults to packer-{{uuid}}._
https://www.packer.io/docs/builders/googlecompute

syuu1228:
On GCE, instance name must be unique. However, "scylla-{{user `scylla_version`| clean_resource_name}}" can be conflict, because we may build multiple images from single Scylla version.
We should use unique instance name, and default value of instance_name is unique since it uses uuid (packer-{{uuid}}).
So let's drop the variable from scylla_gce.json, just apply default value.

Note that this is not about image name, it is temporary instance name for packer.

Fixes #184

### DIFF
--- a/gce/image/scylla_gce.json
+++ b/gce/image/scylla_gce.json
@@ -12,7 +12,6 @@
       "metadata": {"block-project-ssh-keys": "TRUE"},
       "image_family": "scylla",
       "image_name": "scylla-{{user `scylla_version`| clean_resource_name}}-build-{{user `scylla_build_id`| clean_resource_name}}",
-      "instance_name": "scylla-{{user `scylla_version`| clean_resource_name}}",
       "image_description": "Official ScyllaDB image v-{{user `scylla_version`| clean_resource_name}}",
       "use_internal_ip": false,
       "preemptible": true,


yaronkaikov (approved):
LGTM

PR #187, opened 2021-08-16
drop "-force -on-error=abort" option from packer build

syuu1228:
-force -on-error=abort means keep alive packer instance when building
failed, it's good feature to debug images, but if we do that by default
we may leave multiple instances unexpectedly, especially with Jenkins.

It should safer to build images without these options.

Related #184


yaronkaikov (approved):

hagitsegev:
LGTM

PR #188, opened 2021-08-18
On Docker image, make packer always up-to-date

syuu1228:
Packer now has official .rpm/.deb package, let's switch to it and use
up-to-date version to build our EC2/GCE/Azure images.


yaronkaikov (approved):
LGTM

PR #198, opened 2021-09-03
move CPU scaling setup from scylla_image_setup to scylla_install_image

syuu1228:
Currently, CPU scaling setup is mistakenly running in AMI building time.
This is not correct because the AMI image can run other instance type.
To fix this issue, disable CPU scaling setup on scylla_image_setup, and
add it to scylla_install_image.

Fixes scylladb/scylla#9273

### DIFF
--- a/common/scylla_image_setup
+++ b/common/scylla_image_setup
@@ -28,6 +28,7 @@ if __name__ == '__main__':
         print('Failed to initialize RAID volume, exiting setup')
         sys.exit(1)
 
+    run('/opt/scylladb/scripts/scylla_cpuscaling_setup', shell=True, check=True)
     run('/opt/scylladb/scripts/scylla_sysconfig_setup --nic eth0 --setup-nic --ami', shell=True, check=True)
     if cloud_instance.is_supported_instance_class():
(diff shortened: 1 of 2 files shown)


syuu1228:
Note:

I described root cause of scylladb/scylla#9273 at: https://github.com/scylladb/scylla/issues/9273#issuecomment-910177696
Running CPU scaling setup on AMI building time is always wrong, but the bug looks like fixed on latest AMI.
This is because of #9191, before fixed it scylla_cpuscaling_setup always returns 'This computer doesn't supported CPU scaling configuration' on packer temporary instance.
After we fixed it, it become setup CPU scaling correctly.

But as I says earlier, running CPU scaling setup is always wrong, this PR is still needed.

pveentjer (approved):

syuu1228:
@penberg  scylladb/scylla#9193(https://github.com/scylladb/scylla/commit/e5bb88b69a6b2083b7ddb4bede33ac2f8b8c5297) is backported to 4.3, 4.4, 4.5, please backport this as well, since we requires both patches to fix scylladb/scylla#9273

PR #199, opened 2021-09-04
support run build_image.sh/build_deb_image.sh directly

### LINKED ISSUE #197: packer/scripts/build_deb_image.sh - enable using without the old soft link scripts
On https://github.com/scylladb/scylla-machine-image/commit/a7eac3d48bd3791d6f4cb6b24648e287248bbfb1 Takuya "merged" a few similar scripts that use packer to buile images (AMI, GCE, Azure) into one.

I want to set pkg to use the merged script.
This requires also some work on machine-image side.
Currently there are "soft links" instead of the old scripts.
These soft links assume that you run under their default dir, and they run the merged script.
The merged script uses 
PDIRNAME=$(basename $(realpath -se $DIR/..))
to determine the image type.
Add an option (or replace this) with a parameter of the image type.

syuu1228:
Support run build_image.sh/build_deb_image.sh directly by --target <cloud> option.
The script currently located subdirectory of packer to adjust directory
depth of each cloud directories, but it's inconvenient to call longer
path.
So move the script to packer directory, and detect executed via symlink
or not.
To keep compatibility with existing build, direct run will be chdir to
target cloud directory, to use files/ and variables.json for the target.

Fixes #197

### DIFF
--- a/aws/ami/build_ami.sh
+++ b/aws/ami/build_ami.sh
@@ -1,1 +1,1 @@
-../../packer/scripts/build_image.sh
+../../packer/build_image.sh
--- a/aws/ami/build_deb_ami.sh
+++ b/aws/ami/build_deb_ami.sh
@@ -1,1 +1,1 @@
-../../packer/scripts/build_deb_image.sh
+../../packer/build_deb_image.sh
--- a/azure/image/build_azure_image.sh
+++ b/azure/image/build_azure_image.sh
@@ -1,1 +1,1 @@
-../../packer/scripts/build_deb_image.sh
+../../packer/build_deb_image.sh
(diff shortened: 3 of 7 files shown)


hagitsegev (approved):
LGTM, Need to test it with a pkg PR.

syuu1228:
> LGTM, Need to test it with a pkg PR.
> Please write that it is a fix of #197

Updated commit message

hagitsegev:
Testing just to see that current code pkg works (using the soft links):
https://jenkins.scylladb.com/view/nexts/job/scylla-master/job/releng-testing/job/gce-image/446/
So we can merge it.
I'll implement and test direct calls when I'm back from my vacation on https://github.com/scylladb/scylla-pkg/issues/2374

hagitsegev (approved):
LGTM

yaronkaikov (approved):

PR #201, opened 2021-09-07
Revert workaround of kernel-ml boot up issue, fix the problem

syuu1228:
Seem like kernel-ml 5.14.1 boot up issue was caused because xen-blkfront
is not installed on its initramfs, we need to re-generate it with the
driver.

Revert the workaround since we found a solution.


yaronkaikov (approved):
LGTM

yaronkaikov:
Testing: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/738/ -- **OK**

syuu1228:
I think we need to backport this to all supported branches (at least for older releases which doesn't use Ubuntu AMI), otherwise it won't boot up.

yaronkaikov:
@syuu1228 Can you please create a PR for 2021.1 with this commit?

PR #202, opened 2021-09-09
scylla_image_setup: run scylla_cpuscaling_setup earlier

syuu1228:
We found that scylla_cpuscaling_setup won't execute until RAID
construction finished, and it may take few minites or more on larger
systems.
To minimize initialization time, configure CPU scaling first, then
execute rest of setup.

### DIFF
--- a/common/scylla_image_setup
+++ b/common/scylla_image_setup
@@ -22,13 +22,13 @@ from scylla_util import *
 from subprocess import run
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


pveentjer (approved):

yaronkaikov (approved):

yaronkaikov:
@syuu1228 Can you please prepare the cherry-pick for both 4.4 and 4.5?

Issue #196, opened 2021-09-01
Latest build of CentOS7 AMI cannot login via SSH


syuu1228:
On latest build of CentOS7 AMI could not login via SSH, and getting dracut error on system log.
I got this when I build latest next branch on local environment, and also on jenkins:
https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/717/
Looks like some failure happens on initramfs stage, can be related with kernel-ml package.

```
[    3.337884] FDC 0 is a S82078B
         Starting dracut initqueue hook...
[    3.352106] ena 0000:00:03.0: ENA device version: 0.10
[    3.357469] ena 0000:00:03.0: ENA controller version: 0.0.1 implementation version 1
[    3.365367] scsi host0: ata_piix
[    3.365483] ena 0000:00:03.0: LLQ is not supported Fallback to host mode policy.
[    3.369942] scsi host1: ata_piix
[    3.379698] ata1: PATA max MWDMA2 cmd 0x1f0 ctl 0x3f6 bmdma 0xc100 irq 14
[    3.379973] ena 0000:00:03.0: Elastic Network Adapter (ENA) found at mem 83000000, mac addr 02:d7:dd:bb:92:cb
[    3.385039] ata2: PATA max MWDMA2 cmd 0x170 ctl 0x376 bmdma 0xc108 irq 15
[    3.415675] nvme nvme0: pci function 0000:00:1e.0
[    3.440559] nvme nvme0: 15/0/0 default/read/poll queues
[  128.522397] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[  129.047311] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[  129.563310] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[  130.081302] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[  130.597376] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[  131.116381] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[  131.633322] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[  132.151325] dracut-initqueue[353]: Warning: dracut-initqueue timeout - starting timeout scripts
[...]

PR #206, opened 2021-09-13
aws: add xen-blkfront driver to kernel-ml initramfs

syuu1228:
Seem like kernel-ml 5.14.1 boot up issue was caused because xen-blkfront
is not installed on its initramfs, we need to re-generate it with the
driver.

See #196


yaronkaikov (approved):
LGTM

PR #208, opened 2021-09-20
merge scylla.json into single script

syuu1228:
Packer JSON template can define multiple build targets, we can just
combine three different targets into single scripts.


syuu1228:
Note: `packer build` with multiple build targets can build multiple image in single execution, but we do not support it for now.
(we limit target by `packer build --only=$TARGET`)
To support it, we at least need to rename `region` variable, we used same variable name for all targets.

yaronkaikov (approved):
LGTM

PR #209, opened 2021-09-20
merge build_with_docker.sh scripts into single script

syuu1228:
Merge 'build_with_docker.sh' scripts into single script, since it's
almost same between clouds.


packer/build_with_docker.sh
```diff
@@ -0,0 +1,66 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+REALDIR=$(dirname $(readlink -f "$0"))
+DIR=$(dirname $(realpath -se $0))
+PDIRNAME=$(basename $(realpath -se $DIR/..))
+
+if [ "$PDIRNAME" = "aws" ] || [ "$PDIRNAME" = "gce" ] || [ "$PDIRNAME" = "azure" ]; then
```
yaronkaikov, line 21: `build_with_docker.sh` reffer to building based on Centos. we don't support Azure based on at from day one. So it can be drop.

yaronkaikov (changes requested):

yaronkaikov:
@syuu1228 ping

packer/build_with_docker.sh
syuu1228: @yaronkaikov Dropped azure from build_with_docker.sh.

yaronkaikov (approved):

PR #219, opened 2021-09-29
cloudformation: remove ipv6 from ingress security group

### LINKED ISSUE #211: [AWS Cloudformation] Security Groups found with CIDR open to world on ingress
Security Groups found with CIDR open to world on ingress for IPv6. Does this need to be opened to the world? Is there no use case where the customer wants to control this? Even if it needs to be opened to the world, please parameterize it and don't set a default parameter value. In the parameter description, ask the customer to specify CIDR that they desire. Let the customer have more control and be secure by default.

```
  SGAdmin:
    Type: 'AWS::EC2::SecurityGroup'
    Properties:
      GroupDescription: Security group for the admin
      Tags:
        - Key: Name
          Value: !Sub '${ScyllaClusterName}-SGAdmin'
        - Key: ScyllaClusterName
          Value: !Ref ScyllaClusterName
      SecurityGroupIngress:
        - CidrIp: !Ref PublicAccessCIDR
          FromPort: 22
          ToPort: 22
          IpProtocol: tcp
        - CidrIpv6: ::/0
          FromPort: 22
          ToPort: 22
          IpProtocol: tcp
      SecurityGroupEgress:
        - CidrIp: 0.0.0.0/0
          IpProtocol: '-1'
      VpcId: !Ref VPC
```

benipeled:
We don't allow ipv6 on AWS::EC2::VPC so it doesn't affect and relevant
to anything. in addition we don't want to add this option since it's not
common and will complex the flow.

Fixes #211

### DIFF
--- a/aws/cloudformation/scylla.yaml.j2
+++ b/aws/cloudformation/scylla.yaml.j2
@@ -314,10 +314,6 @@ Resources:
           FromPort: 22
           ToPort: 22
           IpProtocol: tcp
-        - CidrIpv6: ::/0
-          FromPort: 22
-          ToPort: 22
-          IpProtocol: tcp
       SecurityGroupEgress:
         - CidrIp: 0.0.0.0/0
           IpProtocol: '-1'


yaronkaikov (approved):
LGTM

Issue #213, opened 2021-09-27
[AWS Cloudformation] Security Groups ingress with an ipProtocol of -1 found.


gnumoreno:
Security Groups ingress with an ipProtocol of -1 found. Do you need all protocols (tcp, udp, icmp, or icmpv6)? 

```
  SGCluster:
    Type: 'AWS::EC2::SecurityGroup'
    Properties:
      GroupDescription: Security group for the cluster
      Tags:
        - Key: Name
          Value: !Sub '${ScyllaClusterName}-SGCluster'
        - Key: ScyllaClusterName
          Value: !Ref ScyllaClusterName
      SecurityGroupIngress:
        - CidrIp: !Ref CIDR
          IpProtocol: '-1'
      SecurityGroupEgress:
        - CidrIp: 0.0.0.0/0
          IpProtocol: '-1'
      VpcId: !Ref VPC
```

gcarmin:
@yaronkaikov  do we really needs to set "all protocol" in:
`IpProtocol: '-1'`
can we just open tcp/udp?

yaronkaikov:
@gcarmin We can set it for tcp or udp or both. Just let us know which range we should enable, since this is a mandatory value when specifing tcp or udp

gcarmin:
@yarongilor  please follow our official documentation on the require port and protocols:
https://docs.scylladb.com/operating-scylla/admin/#id6

yaronkaikov:
@gcarmin You keep tagging the wrong Yaron ;-)

yaronkaikov:
> @yarongilor please follow our official documentation on the require port and protocols: https://docs.scylladb.com/operating-scylla/admin/#id6

Ok, so based on that we only need to enable tcp

PR #222, opened 2021-10-04
SGCluster:set ingress rules to match Scylla ports

yaronkaikov:
Currently SGCluster is open inside for all protocols and ports. This
security group is being used by Scylla internal cluster communication,
let's make sure we only enable relevant tcp ports based on https://docs.scylladb.com/operating-scylla/admin/#id6

Fixes: https://github.com/scylladb/scylla-machine-image/issues/213

### DIFF
--- a/aws/cloudformation/scylla.yaml.j2
+++ b/aws/cloudformation/scylla.yaml.j2
@@ -281,6 +281,7 @@ Resources:
           GroupSet:
             - !Ref SGCluster
             - !Ref SGAdmin
+            - !Ref SGExternal
 {%- endfor %}
   Route:
     Type: 'AWS::EC2::Route'
@@ -297,8 +298,22 @@ Resources:
           Value: !Sub '${ScyllaClusterName}-RT'
         - Key: ScyllaClusterName
           Value: !Ref ScyllaClusterName
-
       VpcId: !Ref VPC
+
+  SGExternal:
+    Type: 'AWS::EC2::SecurityGroup'
+    Properties:
+      GroupDescription: Security group for outbound rules
+      Tags:
+        - Key: Name
+          Value: !Sub '${ScyllaClusterName}-SGExternal'
+        - Key: ScyllaClusterName
+          Value: !Ref ScyllaClusterName
+      SecurityGroupEgress:
+        - CidrIp: 0.0.0.0/0
+          IpProtocol: '-1'
+      VpcId: !Ref VPC
+
   SGAdmin:
     Type: 'AWS::EC2::SecurityGroup'
     Properties:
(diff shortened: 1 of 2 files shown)


benipeled:
Please enable ingress ICMP as well by:
```yaml
- IpProtocol: icmp
  FromPort: 8
  ToPort: -1
  CidrIp: !Ref CIDR
```
In addition:
-  I see no reason for the `SGUser` section - please double check and remove it.
- the `SecurityGroupEgress` is duplicated (`SGCluster` & `SGAdmin`) we might want to take it out for a dedicated one `SGExternal` - WDYT?

yaronkaikov:
> Please enable ingress ICMP as well by:
> 
> ```yaml
> - IpProtocol: icmp
>   FromPort: 8
>   ToPort: -1
>   CidrIp: !Ref CIDR
> ```
> 
> In addition:
> 
> * I see no reason for the `SGUser` section - please double check and remove it.
> * the `SecurityGroupEgress` is duplicated (`SGCluster` & `SGAdmin`) we might want to take it out for a dedicated one `SGExternal` - WDYT?

It looks like `SGUser` is not being used, so i will remove it.

yaronkaikov:
@benipeled regarding the `SecurityGroupEgress`, i think that for `SGCluster` we are currently setting it the wrong way. we should specify `CIDR` as we do for `SecurityGRoupIngress`

benipeled:
> @benipeled regarding the `SecurityGroupEgress`, i think that for `SGCluster` we are currently setting it the wrong way. we should specify `CIDR` as we do for `SecurityGRoupIngress`

`SecurityGRoupIngress` is about inbound traffic so we use the CIDR provides by the user which is the user's private subnet,
but `SecurityGroupEgress` is about outbound traffic which I'm not sure we care about,
Maybe we can remove it completely but here again we complex the flow, if user will try to import something from internal (or any other outbound communication) he'll have to deal with network "issues" and may give up.

gcarmin (approved):
LGTM

aws/cloudformation/scylla.yaml.j2
```diff
@@ -297,8 +298,22 @@ Resources:
           Value: !Sub '${ScyllaClusterName}-RT'
         - Key: ScyllaClusterName
           Value: !Ref ScyllaClusterName
+      VpcId: !Ref VPC
 
+  SGExternal:
+    Type: 'AWS::EC2::SecurityGroup'
+    Properties:
+      GroupDescription: Security group for the admin
```
benipeled, line 306: "Security group for outbound rules"

benipeled (approved):
Require one tiny fix but LGTM

aws/cloudformation/scylla.yaml.j2
yaronkaikov: Fixed,. Thanks

bentsi (approved):

PR #224, opened 2021-10-05
packer/build_deb_image: Improve error handling

hagitsegev:
- Fix success verification
- Error when missing target parameter


hagitsegev:
Tested: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/gce-image/479/ **OK**
I also tested it with missing target parameter (but build was already erased).

packer/build_deb_image.sh
```diff
@@ -292,14 +295,22 @@ export PACKER_LOG_PATH
...
   elif [ "$TARGET" = "gce" ]; then
       grep "A disk image was created" $PACKER_LOG_PATH
+      GREP_STATUS=$?
   elif [ "$TARGET" = "azure" ]; then
     grep "Builds finished. The artifacts of successful builds are:" $PACKER_LOG_PATH
+    GREP_STATUS=$?
+  else
+    echo "No Target is defined"
+    GREP_STATUS=1
   fi
```
yaronkaikov, line 311: Instead of those if else, let's make it more clear and rubust. one why to do it:
```
 case "$TARGET" in
      "aws")
        grep "us-east-1:" $PACKER_LOG_PATH
        ;;
      "gce")
        grep "A disk image was created" $PACKER_LOG_PATH
        ;;
      "azure")
       grep "Builds finished. The artifacts of successful builds are:" $PACKER_LOG_PATH
       ;;
      *)
        echo "No Target is defined"
        exit 1
    esac
```

packer/build_deb_image.sh
```diff
@@ -292,14 +295,22 @@ export PACKER_LOG_PATH
...
-  if [ $? -ne 0 ] ; then
+  echo "GREP_STATUS: |$GREP_STATUS|"
```
yaronkaikov, line 312: You don't need the `GREP_STATUS=1` definition in line 298
instead you can do:
```
if [ $? -ne 0 ] ; then
   echo "Say something"
   exit 1
else
   echo "say something else"
   exit 0
fi

**Review by yaronkaikov — CHANGES_REQUESTED** [review]

File: packer/build_deb_image.sh (hunk 2 above)
Comment (hagitsegev) [reply packer/build_deb_image.sh:312]: In general it is better to have a single exit point.
There are a few exit points in this script, but I don't want to add more.

File: packer/build_deb_image.sh [hunk 3]
```diff
@@ -292,14 +295,26 @@ export PACKER_LOG_PATH
...
+      ;;
+    "gce")
       grep "A disk image was created" $PACKER_LOG_PATH
-  elif [ "$TARGET" = "azure" ]; then
-    grep "Builds finished. The artifacts of successful builds are:" $PACKER_LOG_PATH
-  fi
-  if [ $? -ne 0 ] ; then
+      GREP_STATUS=$?
+      ;;
+    "azure")
```
yaronkaikov, line 308: indentation

yaronkaikov (changes requested):

hagitsegev:
@syuu1228 / @yaronkaikov Please re-review.

yaronkaikov (changes requested):

yaronkaikov (approved):

PR #225, opened 2021-10-06
Revert adding 'name' option for builders

syuu1228:
On 4f5c2a7e4d0f8c8da8bf2630bcfe71dad07c48bb, we added 'name' option for
each IaaS builders, to pass target cloud name for scylla_install_image.
However, this change affect output of packer.log, it renamed builder
name on the log like this.

Before: googlecompute: Disk has been deleted!
After: gce: Disk has been deleted!

We are checking log output on Jenkins to make sure the build suceed,
so we shouldn't change the output.

### DIFF
--- a/packer/build_deb_image.sh
+++ b/packer/build_deb_image.sh
@@ -130,6 +130,14 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+if [ "$TARGET" = "aws" ]; then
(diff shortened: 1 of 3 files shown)


hagitsegev:
There is no need to revert this.
We had this simple https://github.com/scylladb/scylla-pkg/pull/2425 to over come this change on pkg.
Unless you see another reason to undo it.

PR #233, opened 2021-10-12
GCE: use underscores in image labels

enaydanov:
Unfortunately, GCE API works pretty bad with dashes in labels.
We can't align with AWS here and switched to underscores.


bentsi (approved):

hagitsegev:
LOL
We had this long discussion on dashes and underscores...

Issue #214, opened 2021-09-27
[AWS Cloudformation] Caution: password authentication isn't enabled by default


gnumoreno:
"Caution: password authentication isn't enabled by default". Can you activate the password per default and set it to the instance ID? You can get the instance id from the metadata. Take a look at the following link and just add instance-id at the end of the meta-data url. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instance-metadata-ex-2

yaronkaikov:
@gcarmin Where did you guys see this error?

PR #237, opened 2021-10-19
cloudformation: reset the deafult password for cassandra user

benipeled:
Add an option to run commands after scylla-server is up and running

We have a request to make sure the password for the default user
(cassandra) will be changed during cluster creation,

Since Scylla has no setup-flag for changing password during setup, we
need to do it by running the "ALTER USER..." commamd,

The image-setup mechanism built to run before scylla-server service is
started, which means we can't use it for running the ALTER command,

This commit adds a new systemd service `scylla-image-post-start` for
running commands after scylla-server is up and running. it's a copy of
the mechanism use by `scylla-image-setup` which is retrieving a key from
instance `UserData`. Ths service is running once during the initial
boot.

Following the post_start_script option added on 57575d0, this commit
makes sure the cassandra' user password will be change to the
instance-id of the first node.

Fix: #214

### DIFF
--- a/aws/cloudformation/scylla.yaml.j2
+++ b/aws/cloudformation/scylla.yaml.j2
@@ -240,6 +240,7 @@ Resources:
                     #!/bin/bash -ex
                     export INSTANCE_ID=$(curl -sS http://169.254.169.254/latest/meta-data/instance-id)
                     /usr/bin/cqlsh -u cassandra -p cassandra -e "ALTER USER cassandra WITH PASSWORD '$INSTANCE_ID';"
+                    /usr/bin/cqlsh -u cassandra -p $INSTANCE_ID -e "ALTER KEYSPACE system_auth WITH REPLICATION = {'class': 'SimpleStrategy', 'replication_factor': ${InstanceCount}};"
             - '"}'
 {%- else %}
       UserData: !Base64


packer/scylla_install_image
```diff
@@ -147,6 +147,8 @@ if __name__ == '__main__':
 
     run('systemctl daemon-reload')
     run('systemctl enable scylla-image-setup.service')
+    if distro == 'ubuntu':
```
syuu1228, line 150: I want to avoid `if distro` here, if it possible.
We may not use CentOS AMI on newer releases but this change still can work on CentOS correct?
If it possible, please drop `if distro` line and also add files to rpm too.

common/scylla-image-post-start.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Scylla cloud image post start service
+After=scylla-server.service
+ConditionPathExists=!/etc/scylla/machine_image_post_start
```
syuu1228, line 4: Not really important, but on scylla-image-setup.service the file is `/etc/scylla/machine_image_configured`, so let's align this one to `/etc/scylla/machine_image_post_start_configured`.

common/scylla-image-post-start.service
```diff
@@ -0,0 +1,16 @@
...
+
+[Service]
+Type=oneshot
+# The ExecStartPre makes sure the scylla_post_start.py script 
+# will run only after scylla-server is up and running
+ExecStartPre=/bin/sleep 30
+ExecStart=/opt/scylladb/scylla-machine-image/scylla_post_start.py
```
syuu1228, line 11: Do we required root privilege for post_start script?
If not, it's better to use `User=scylla` to run the script as a non-root user.
It feels like safer since we run shellscript received from external place.

common/scylla-image-post-start.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Scylla cloud image post start service
+After=scylla-server.service
+ConditionPathExists=!/etc/scylla/machine_image_post_start
+
+[Service]
+Type=oneshot
+# The ExecStartPre makes sure the scylla_post_start.py script 
+# will run only after scylla-server is up and running
+ExecStartPre=/bin/sleep 30
```
syuu1228, line 10: I tried to find better way to do this (in systemd directive), but not really able to find. There is a way to do same using .timer, I want to avoid to do that for just waiting few sec.
Anyway, did you tried to run script without this delay? It didn't work?
If so, we might have a bug on scylla-server side, it should be able to connect when it becomes 'running' state.
But it is another issue, you can keep this line if it needed.

packer/scylla_install_image
benipeled: Currently it's not gonna work since I didn't add the service to the RPM version,
I'll try adding it and remove this condition.

slivne:
Not a review comment - but checking - how do you validate the whole cluster is up and not only a single node / subset ?

will this service be run on all the nodes or only one ?

PR #244, opened 2021-11-13
scylla_image_setup: add check current instance has ephemeral disks

### LINKED ISSUE #243: GCE: GCE image can use persistent disks without any warning message
On AWS, when we launch our AMI with EBS disks only, it will show warning message on login prompt, and don't run scylla_io_setup automatically.
The warning message says:
```
$INSTANCE_TYPE is not eligible for optimized automatic tuning!
To continue startup ScyllaDB on this instance, you need to attach additional disks, then run 'sudo scylla_io_setup' then 'systemctl start scylla-server'.
For a list of optimized instance types and more EC2 instructions see '$URL'
```
It is because the configuration is not optimal one, and iotune may takes 5-10min to complete, if we run automatically it will hard to understand why scylla-server.service does not start.

However, on GCE image behavior is different.
If we run our image in "supported" instance type and attach persistent disks, it does not print any warning message on login prompt, and silently run iotune.

It is because difference between AWS and GCE.
On AWS, "supported" instance type forced to attach ephemeral disks, so EBS only configuration means "unsupported" instance type.
So scylla_image_setup does:
```
    if cloud_instance.is_supported_instance_class():
        cloud_instance.io_setup()
```
However, GCE does not force to attach ephemeral disks on "supported" instance.
Which means we also need to check ephemeral disks, run `cloud_instance.io_setup` only when ephemeral disks are available.

syuu1228:
On AWS, "supported" instance type means it should have ephemeral disks.
So scylla_image_setup checks instance type before running io_setup like
this:
    if cloud_instance.is_supported_instance_class():
        cloud_instance.io_setup()

But this is not enough for GCE, it can configure instance without
ephemeral disks even instance type is 'supported' one.
To support such case, we also need to check the instance has ephemeral disks.

Fixes #243

Signed-off-by: Takuya ASADA <syuu@scylladb.com>

### DIFF
--- a/common/scylla_image_setup
+++ b/common/scylla_image_setup
@@ -33,7 +33,13 @@ if __name__ == '__main__':
 
     run('/opt/scylladb/scripts/scylla_sysconfig_setup --nic eth0 --setup-nic --ami', shell=True, check=True)
     if cloud_instance.is_supported_instance_class():
-        cloud_instance.io_setup()
+        # We run io_setup only when ehpemeral disks are available
+        if is_gce():
+            nr_disks = cloud_instance.nvmeDiskCount
+            if nr_disks < 0:
+                cloud_instance.io_setup()
+        else:
+            cloud_instance.io_setup()
     if os.path.ismount('/var/lib/scylla'):
         run('/opt/scylladb/scripts/scylla_coredump_setup --dump-to-raiddir', shell=True, check=True)
     else:


yaronkaikov:
@syuu1228 Please rebase your branch. @bentsi sent a fix for the workflow which is currently failing

syuu1228:
@yaronkaikov rebased with latest next, but no conflict.

yaronkaikov (approved):

PR #260, opened 2021-12-02
Build ami scripts: Make default product really take effect

eliransin:
In the current scripts, if product is not given, even though it
is defaulted to whatever is in `build/SCYLLA-PRODUCT-FILE`, it
will not propagate to the packer vars (through `--install-args`).
This is a problem if the default is different than the one set in
`packer/scylla_install_image`.
This change will make the default always propagate to the packer
whether it is the default or a user set value.


yaronkaikov (approved):
LGTM - Thanks

PR #263, opened 2021-12-12
AWS AMI: add `ami-regions` parameter control

eliransin:
The AMI builder of the packer can copy the newly created AMI as
part of the normal flow. This has the advantage of copying the
image along with the Tags and Permissions set without having to
have a dedicated script for doing so.
Here we add an optional `--ami-regions` parameter that takes a
comma separated list of regions and copy the AMI once it's build
has succeeded.

Signed-off-by: Eliran Sinvani <eliransin@scylladb.com>


yaronkaikov:
Tested https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/977/ - **OK**

yaronkaikov (approved):
LGTM - Thanks @eliransin

Issue #265, opened 2021-12-13
Add support of --raid-level parameter


aleksbykov:
Commit  https://github.com/scylladb/scylla/commit/42fd73d0332d034809c6cd228785bff82e1845aa added support to create
raid 5 by passing parameter --raid-leve to scylla_setup and scylla_raid_setup.

Need add support SMI scripts to create different raid configuration with aws ami and gce images

yaronkaikov:
Closed via https://github.com/scylladb/scylla-machine-image/pull/266

PR #266, opened 2021-12-13
scylla_create_device: Add support --raid-level parameter

aleksbykov:
With commit https://github.com/scylladb/scylla/commit/42fd73d0332d034809c6cd228785bff82e1845aa
support RAID 5 was added to scylla.
Added appropriate parameter to scylla_configure and
scylla_create_device --raid-level parameter
for configure scylla ami

Issue: https://github.com/scylladb/scylla-machine-image/issues/265

### DIFF
--- a/common/scylla_configure.py
+++ b/common/scylla_configure.py
@@ -48,15 +48,16 @@ class ScyllaMachineImageConfigurator:
         'post_configuration_script': '',
         'post_configuration_script_timeout': 600,  # seconds
         'start_scylla_on_first_boot': True,
-        'data_device': 'auto'  # Supported options:
-                               #   instance_store - find all ephemeral devices (only for AWS)
-                               #   attached - find all attached devices and use them
-                               #   auto - automatically select devices using following strategy:
-                               #       GCE: select attached NVMe.
-                               #       AWS:
-                               #           if ephemeral found - use them
-                               #           else if attached EBS found use them
-                               #           else: fail create_devices
+        'data_device': 'auto',  # Supported options:
+                                #   instance_store - find all ephemeral devices (only for AWS)
+                                #   attached - find all attached devices and use them
+                                #   auto - automatically select devices using following strategy:
+                                #       GCE: select attached NVMe.
+                                #       AWS:
+                                #           if ephemeral found - use them
+                                #           else if attached EBS found use them
+                                #           else: fail create_devices
+        'raid_level': 0  # Default raid level is 0, supported raid 0, 5
     }
 
     DISABLE_START_FILE_PATH = Path("/etc/scylla/ami_disabled")
@@ -163,9 +164,11 @@ class ScyllaMachineImageConfigurator:
 
     def create_devices(self):
         device_type = self.instance_user_data.get("data_device", self.CONF_DEFAULTS['data_device'])
+        raid_level = self.instance_user_data.get("raid_level", self.CONF_DEFAULTS['raid_level'])
+        cmd_create_devices = f"/opt/scylladb/scylla-machine-image/scylla_create_devices --data-device {device_type} --raid-level {raid_level}"
         try:
             LOGGER.info(f"Create scylla data devices as {device_type}")
-            subprocess.run(f"/opt/scylladb/scylla-machine-image/scylla_create_devices --data-device {device_type}", shell=True, check=True)
+            subprocess.run(cmd_create_devices, shell=True, check=True)
         except Exception as e:
             LOGGER.error("Failed to create devices: %s", e)
 
--- a/common/scylla_create_devices
+++ b/common/scylla_create_devices
@@ -38,7 +38,7 @@ class NoDiskFoundError(Exception):
     def __str__(self):
         return "No data disk found, abort setup"
 
-def create_raid(devices):
+def create_raid(devices, raid_level=0):
     if len(devices) == 0:
         raise(NoDiskFoundError)
     scylla_path = Path("/var/lib/scylla")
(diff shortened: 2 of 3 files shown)


fruch (dismissed):
LGTM

roydahan:
@slivne to decide if it goes into 4.6 or not.

yaronkaikov:
@aleksbykov can you share your testing ? did you verify both raid 0 and 5 are working?

roydahan:
According to @slivne it won't be part of 4.6, so no backport is needed.

yaronkaikov (approved):

common/scylla_configure.py
```diff
@@ -161,11 +162,16 @@ def start_scylla_on_first_boot(self):
             LOGGER.info("Disabling Scylla start on first boot")
             self.DISABLE_START_FILE_PATH.touch()
 
-    def create_devices(self):
+    def create_devices(self, dry_run=False):
         device_type = self.instance_user_data.get("data_device", self.CONF_DEFAULTS['data_device'])
+        raid_level = self.instance_user_data.get("raid_level", self.CONF_DEFAULTS['raid_level'])
+        cmd_create_device = f"/opt/scylladb/scylla-machine-image/scylla_create_devices --data-device {device_type} --raid-level {raid_level}"
```
bentsi, line 168: `cmd_create_device` - > `cmd_create_devices`

common/scylla_configure.py
```diff
@@ -161,11 +162,16 @@ def start_scylla_on_first_boot(self):
             LOGGER.info("Disabling Scylla start on first boot")
             self.DISABLE_START_FILE_PATH.touch()
 
-    def create_devices(self):
+    def create_devices(self, dry_run=False):
```
bentsi, line 165: As far as I understand `dry_run` arg is created only for testing.
In such case you should use `unittest.mock.patch` in the tests. See example below

tests/test_scylla_configure.py
```diff
@@ -178,3 +178,20 @@ def test_do_not_start_on_first_boot(self):
...
+
+    def test_default_raid0(self):
+        self.run_scylla_configure(**self.default_instance_metadata())
+        actual_cmd = self.configurator.create_devices(dry_run=True)
+        assert "--raid-level 0" in actual_cmd
+
+    def test_set_raid0(self):
+        raw_user_data = json.dumps(dict(raid_level=0))
+        self.run_scylla_configure(user_data=raw_user_data, private_ipv4=self.private_ip)
+        actual_cmd = self.configurator.create_devices(dry_run=True)
```
bentsi, line 190:
```suggestion
        with unittest.mock.patch("subprocess.run") as mocked_run:
            self.configurator.create_devices()
            actual_cmd = mocked_run.call_args[0][0]
```

common/scylla_configure.py
```diff
@@ -48,15 +48,16 @@ class ScyllaMachineImageConfigurator:
...
+        'data_device': 'auto',  # Supported options:
+                                #   instance_store - find all ephemeral devices (only for AWS)
+                                #   attached - find all attached devices and use them
+                                #   auto - automatically select devices using following strategy:
+                                #       GCE: select attached NVMe.
+                                #       AWS:
+                                #           if ephemeral found - use them
+                                #           else if attached EBS found use them
+                                #           else: fail create_devices
+        'raid_level': '0'  # Default raid level is 0, supported raid 0, 5
```
bentsi, line 60: let's be consistent, use `int` for integers

common/scylla_create_devices
```diff
@@ -104,6 +107,9 @@ if __name__ == "__main__":
     parser.add_argument('--data-device', dest='data_device', action='store',
                         choices=["auto", "attached", "instance_store"],
                         help='Define type of device to use for scylla data: attached|instance_store')
+    parser.add_argument('--raid-level', dest='raid_level', action='store',
+                        choices=["0", "5"],
```
bentsi, line 111: type should be int, you check in line 44 integers

common/scylla_create_devices
```diff
@@ -38,16 +38,19 @@ class NoDiskFoundError(Exception):
     def __str__(self):
         return "No data disk found, abort setup"
 
-def create_raid(devices):
+def create_raid(devices, raid_level):
     if len(devices) == 0:
         raise(NoDiskFoundError)
+    if raid_level not in [0, 5]:
```
bentsi, line 44: the check is actually done in the argument parser, no need to double-check it

common/scylla_create_devices
bentsi: anyway, the code in lines 44-46 is redundant, so for consistency matters use int or str

bentsi (changes requested):
please address comments

common/scylla_create_devices
bentsi: also set default there

common/scylla_configure.py
aleksbykov: done

tests/test_scylla_configure.py
aleksbykov: Done

common/scylla_create_devices
aleksbykov: fixed

common/scylla_create_devices
aleksbykov: fixed

aleksbykov:
@bentsi could you take a look again

common/scylla_create_devices
```diff
@@ -47,7 +47,7 @@ def create_raid(devices):
         print(f"{scylla_path} is already mounted. Will not run 'scylla_raid_setup'!")
         sys.exit(0)
     run(["/opt/scylladb/scripts/scylla_raid_setup", "--raiddev", "/dev/md0", "--disks", ",".join(devices),
-         "--root", "/var/lib/scylla", "--volume-role", "all", "--update-fstab"], check=True)
+         "--root", "/var/lib/scylla", "--raid-level", raid_level, "--volume-role", "all", "--update-fstab"], check=True)
```
bentsi, line 50: if you call run without `shell=True` you need to pass list of strings
```suggestion
         "--root", "/var/lib/scylla", "--raid-level", str(raid_level), "--volume-role", "all", "--update-fstab"], check=True)
```
otherwise, you will get Exception `TypeError: expected str, bytes or os.PathLike object, not int`:
image

aleksbykov: fixed

bentsi (approved):

PR #272, opened 2021-12-20
gce: increase disk size to 30GB

syuu1228:
Since we configure almost same settings on all IaaS images, we should
use same default root disk size on all images.

However, only GCE image specifies 20GB disk size, we should increase it
to 30GB (AWS and Azure default).

Note: we don't set default size on Azure build, but default disk size on
Azure is 30GB.


yaronkaikov (approved):
LGTM

PR #273, opened 2021-12-21
[ami]:Add to scylla-packages manifast file arch prefix

yaronkaikov:
Since today we are creating AMI for both `x86` and `arm` we should have
2 manifast files to upload.

Adding arch prefix so they will not override each other

This is a preparation for https://github.com/scylladb/scylla-pkg/issues/1869


syuu1228 (approved):
LGTM.

bentsi (approved):

PR #276, opened 2021-12-22
fix(build_deb_image.sh): set arch var only for AWS

yaronkaikov:
In 63ab4ae3a1e950a5c32fd8b7c532772b8bb76b25 i have added another variable to packer with the arch type. By mistake it was added to all cloud providers rather then only AWS

Fixing it


fruch (approved):
LGTM

PR #278, opened 2021-12-27
scylla_configure.py: support Ec2MultiRegionSnitch

syuu1228:
To support Ec2MultiRegionSnitch on EC2, we need to use public ip for
broadcast_address and seeds.

See scylladb/scylla#9386

### DIFF
--- a/common/scylla_configure.py
+++ b/common/scylla_configure.py
@@ -107,10 +107,22 @@ class ScyllaMachineImageConfigurator:
         return self._instance_user_data
 
     def updated_ami_conf_defaults(self):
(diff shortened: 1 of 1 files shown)


syuu1228:
This is scylla-machine-image part of https://github.com/scylladb/scylla/pull/9846

fruch:
@syuu1228 FYI you are breaking the unittests

syuu1228:
Closing, since we haven't find a way to avoid compatibility issue with existing (older) versions of our AMIs:
https://github.com/scylladb/scylla/issues/9386#issuecomment-1001531061

PR #279, opened 2021-12-30
aws: find latest image for source AMI

syuu1228:
We currently hardcoded source AMI ID, it may outdated if we keep using old one.
To automatically use latest image for source AMI, drop source_ami and
switch to source_ami_filter to find latest image by name pattern and owner id.

### DIFF
--- a/packer/build_deb_image.sh
+++ b/packer/build_deb_image.sh
@@ -272,16 +272,17 @@ SCYLLA_VERSION=$(echo $SCYLLA_VERSION | sed 's/\(.*\)\~)*/\1./')
 
 if [ "$TARGET" = "aws" ]; then
     SSH_USERNAME=ubuntu
-    declare -A AMI
-    AMI=(["x86_64"]=ami-04505e74c0741db8d ["aarch64"]=ami-0ae74ae9c43584639)
+    SOURCE_AMI_OWNER=099720109477
     REGION=us-east-1
 
     arch="$(uname -m)"
     case "$arch" in
       "x86_64")
+        SOURCE_AMI_FILTER="ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64*"
         INSTANCE_TYPE="c4.xlarge"
         ;;
       "aarch64")
+        SOURCE_AMI_FILTER="ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64*"
         INSTANCE_TYPE="a1.xlarge"
         ;;
       *)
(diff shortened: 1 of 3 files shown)


benipeled:
1. Does this filter distinguish between dev & GA versions?
2. I'm not sure about auto-using the latest version - imagine a new version with some library-change that affect us by somehow,
It might take us some time to find the root cause (+ a customer who is trying to understand a behavior-change between his nodes) - WDYT?

yaronkaikov:
@benipeled AFAICT it's an official version. today we perform update and upgrade anyway when building new AMI. so i don't think (or hope) it will affect us much. also we always running tests before releasing so i don't see an issue here

@syuu1228 Please rebase since you have conflicts . We should anyway run a test before merging it and make sure we can build and run SCT artifact test

syuu1228:
@yaronkaikov @benipeled When I sent this PR I thought why we have older packages on our AMI is maybe because we use older source AMI (https://github.com/scylladb/scylla-pkg/issues/2648), but it was not correct.
Because we run `yum update` or `apt-get upgrade` on every AMI build (https://github.com/scylladb/scylla-machine-image/blob/next/packer/scylla_install_image#L112).
So this will only accelarate `yum update` or `apt-get upgrade` speed while building new AMI, since the source AMI is up-to-date, and does not change anything else.

syuu1228:
@yaronkaikov Pushed rebase version

syuu1228:
> @yaronkaikov @benipeled When I sent this PR I thought why we have older packages on our AMI is maybe because we use older source AMI ([scylladb/scylla-pkg#2648](https://github.com/scylladb/scylla-pkg/issues/2648)), but it was not correct. Because we run `yum update` or `apt-get upgrade` on every AMI build (https://github.com/scylladb/scylla-machine-image/blob/next/packer/scylla_install_image#L112). So this will only accelarate `yum update` or `apt-get upgrade` speed while building new AMI, since the source AMI is up-to-date, and does not change anything else.

And we already build GCE and Azure images in this way (find latest source image automatically), so this applies same way to build the image on AWS too.

fruch:
@syuu1228 FYI, for the centos 7 it was problematic, since we were getting newer kernel versions (and sometimes broken ones), in some release on each release candidate we got a different kernel, and it was slowing down the testing and release.

I think that for master it might be o.k., but on releases branches we might want to freeze the source images in some way.

yaronkaikov:
@fruch We don't build centos 7 anymore. this change will probably stay in master only (and future releases)

fruch:
> @fruch We don't build centos 7 anymore. this change will probably stay in master only (and future releases)

centos 7 was just an example, where we got bitten from taking latest AMI each time we build.
I'm just saying that for releases branches we might better off with pinning it.

syuu1228:
> > @fruch We don't build centos 7 anymore. this change will probably stay in master only (and future releases)
> 
> centos 7 was just an example, where we got bitten from taking latest AMI each time we build. I'm just saying that for releases branches we might better off with pinning it.

Maybe we can add a logic to choose specified AMI ID or find latest ID on build_deb_image.sh / build_image.sh and generate scylla.json for it (probably requires to templatize scylla.json).

yaronkaikov (approved):
LGTM

PR #280, opened 2021-12-30
fix(scylla_install_image):create manifast based on product name

yaronkaikov:
Following the work in https://github.com/scylladb/scylla-machine-image/pull/97/commits/e2243830305066756138943f291b85565f72709c
manifest creation is failing in enterprise since we use `scylla` as hard coded rather then product name

Fixing it


fruch (approved):
LGTM

PR #281, opened 2022-01-02
scylla_install_image:add Scylla version to manifest filename

yaronkaikov:
Later in the process of building the AMI we need to upload to S3 for future use.
Adding into manifest filename the current Scylla version


fruch (approved):
LGTM

PR #287, opened 2022-01-17
build_deb_image.sh: create image manifest with image ID

yaronkaikov:
Today in scylla-pkg we search the log for the created AMI/Gce/Azure image.
There is an option when building image with packer to auto create a manifest with the relevant information , this will allow us to send the functions the value instead of looking for it


fruch (approved):
LGTM

PR #288, opened 2022-01-17
make scylla-server.service stopped state when scylla_on_first_boot=false

### LINKED ISSUE #234: scylla-server.service fails when start_scylla_on_first_boot param is set to false instead of being stopped
## Brief Description of Issue

**Found in:** Scylla 4.3.2 AMI but I suspect it extends to all generated images

When the `start_scylla_on_first_boot` param is set to `false` in user data what actually happens is [it touches a file](https://github.com/scylladb/scylla-machine-image/blob/scylla-4.3.2/common/scylla_configure.py#L144) which [`scylla_prepare` looks for](https://github.com/scylladb/scylla/blob/scylla-4.3.2/dist/common/scripts/scylla_prepare#L113) when `scylla-server.service` is started and if it finds it then it deletes it and exits with return code 1. While in the failed state, it technically isn't started, but any time the systemd daemon is reloaded or any other service is manipulated (E.g. `systemctl restart rsyslog`) then it automatically tries to start `scylla-server.service` again which will succeed because the marker file gets deleted by `scylla_prepare`.

## Workaround

There does exist a workaround: if you run `systemctl stop scylla-server.service` then Scylla will try to start and then immediately get stopped and will stay in the stopped state.

## Expected behaviour

Personally, if `start_scylla_on_first_boot` is set to `false`, then I'd expect the service to be disabled in a systemd world. This would allow me to perform last minute configuration before starting a cluster or joining the new instance to the cluster.

syuu1228:
Currently, when scylla_on_first_boot set to false, scylla-server.service
becomes failed state by returning non-zero in scylla_prepare script.
But systemd may automatically tries to start scylla-server.service again
when daemon is reloaded or any other services is manipulated.

Instead, we should make scylla-server.service stopped state, by running
systemctl stop scylla-server.service.

Fixes #234

### DIFF
--- a/common/scylla_configure.py
+++ b/common/scylla_configure.py
@@ -60,8 +60,6 @@ class ScyllaMachineImageConfigurator:
         'raid_level': 0  # Default raid level is 0, supported raid 0, 5
     }
 
-    DISABLE_START_FILE_PATH = Path("/etc/scylla/ami_disabled")
-
     def __init__(self, scylla_yaml_path="/etc/scylla/scylla.yaml"):
         self.scylla_yaml_path = Path(scylla_yaml_path)
         self.scylla_yaml_example_path = Path(scylla_yaml_path + ".example")
@@ -159,7 +157,7 @@ class ScyllaMachineImageConfigurator:
         default_start_scylla_on_first_boot = self.CONF_DEFAULTS["start_scylla_on_first_boot"]
         if not self.instance_user_data.get("start_scylla_on_first_boot", default_start_scylla_on_first_boot):
             LOGGER.info("Disabling Scylla start on first boot")
-            self.DISABLE_START_FILE_PATH.touch()
+            subprocess.run("/usr/bin/systemctl stop scylla-server.service", shell=True, check=True)
 
     def create_devices(self):
         device_type = self.instance_user_data.get("data_device", self.CONF_DEFAULTS['data_device'])
--- a/tests/test_scylla_configure.py
+++ b/tests/test_scylla_configure.py
@@ -89,7 +89,6 @@ class TestScyllaConfigurator(TestCase):
     def test_empty_user_data(self):
         self.run_scylla_configure(**self.default_instance_metadata())
         self.check_yaml_files_exist()
-        assert not self.configurator.DISABLE_START_FILE_PATH.exists(), "ami_disabled not created"
         with self.configurator.scylla_yaml_path.open() as scylla_yaml_file:
             LOGGER.info("Checking that defaults are set as expected...")
             scylla_yaml = yaml.load(scylla_yaml_file, Loader=yaml.SafeLoader)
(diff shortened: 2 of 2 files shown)


common/scylla_configure.py
```diff
@@ -160,7 +158,7 @@ def start_scylla_on_first_boot(self):
         default_start_scylla_on_first_boot = self.CONF_DEFAULTS["start_scylla_on_first_boot"]
         if not self.instance_user_data.get("start_scylla_on_first_boot", default_start_scylla_on_first_boot):
             LOGGER.info("Disabling Scylla start on first boot")
-            self.DISABLE_START_FILE_PATH.touch()
```
fruch, line 163: are you sure calling `sysctl stop` would make is not start on next boot ?

fruch: I think we should call `systemctl disable --now` instead to make sure it won't start on next boot

syuu1228: You are right if we want to permanently disable scylla-server.service.
But `start_scylla_on_first_boot=false` sounds like disabling it only on first boot, because it says "on_first_boot".
Which behavior is expected for this option?

fruch: The idea of the feature, is that user would be able to configure the cluster manually.
So first boot, it's a must it would be stopped.
If we disable the service, we need to document this change, that user need to also enable it, not just start it.

Otherwise we can still use the hack with the touched file

fruch: Also check out the unittests, I think they are failing.

syuu1228: > Also check out the unittests, I think they are failing.

Fixed the test failure (removed assert self.configurator.DISABLE_START_FILE_PATH.exists() since we don't add the file anymore).

syuu1228: > The idea of the feature, is that user would be able to configure the cluster manually. So first boot, it's a must it would be stopped. If we disable the service, we need to document this change, that user need to also enable it, not just start it.

So users expected `scylla_on_first_boot=false` means stop scylla on first boot, then auto-start scylla on next boot.
This is what `systemctl stop scylla-server.service` does.
If we change it to `systemctl disable --now`, we need to document to run `systemctl enable scylla-server.service` before reboot, as you described.

> Otherwise we can still use the hack with the touched file

I thought current touched file hack is enough, but #234 reported that systemd tries to restart scylla-server.service again and again, when " the systemd daemon is reloaded or any other service is manipulated (E.g. systemctl restart rsyslog)".
It is probably because scylla-server.service is failure state, not stopped state, so we need to take care for it.

Alternative approach to fix it is, add `ConditionPathExists=!/etc/scylla/ami_disabled` on scylla-server.service.
But it can stop scylla-server.service but requires to remove the touch file before starting scylla-server, so it requires another documentation.

fruch (approved):
LGTM

common/scylla_configure.py
```diff
@@ -60,8 +60,6 @@ class ScyllaMachineImageConfigurator:
         'raid_level': 0  # Default raid level is 0, supported raid 0, 5
     }
 
-    DISABLE_START_FILE_PATH = Path("/etc/scylla/ami_disabled")
```
fruch, line 63: now I've remembered in something else, the cloud team was using this file to wait for scylla machine image finish it's boot.

you'll need to cross check with them if they are still using this file as the indicator, or moved to using a different method

fruch: it's actually a different file:
https://github.com/scylladb/siren/blob/f0c2ffae40618e395ec2a196219a54c6207b4bb2/cluster/server/instance.go#L28

so scratch that...

syuu1228:
Jenkins test passed:
https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/next-machine-image/247/

syuu1228:
ping @yaronkaikov

yaronkaikov (approved):

PR #290, opened 2022-01-24
prepare for Scylla 5.0

yaronkaikov:
We decided to name the next version Scylla 5.0, in honor of
Raft based schema management.


fruch (approved):
LGTM

PR #291, opened 2022-01-25
Fix CentOS image build

syuu1228:
On latest master CentOS image build is broken because we added "arch" and "product" variables on Packer.
To fix it we need to add them to build_image.sh too.


fruch:
@syuu1228  out of curiosity, why did you tried building the centos one ?

fruch (approved):
LGTM

yaronkaikov (approved):

PR #292, opened 2022-01-25
treewide: use Software Package Data Exchange (SPDX) license identifiers

syuu1228:
This is scylla-machine-image version of scylladb/scylla#9937

----

Instead of lengthy blurbs, switch to single-line, machine-readable
standardized (https://spdx.dev) license identifiers. The Linux kernel
switched long ago, so there is strong precedent.

Note that dist/debian/debian_files_gen.py is replaced from AGPL to Apache
since applying AGPL was mistake.

### DIFF
--- a/aws/cloudformation/test_scylla_cfn.py
+++ b/aws/cloudformation/test_scylla_cfn.py
@@ -1,17 +1,7 @@
 #
 # Copyright 2020 ScyllaDB
 #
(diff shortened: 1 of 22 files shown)


fruch (dismissed):
LGTM

syuu1228:
Rebased with latest next, also applied SPDX identifier on newly added scripts.

yaronkaikov (approved):
LGTM

PR #293, opened 2022-01-31
pre-allocate swapfile while building image

### LINKED ISSUE #285: swap creation slows down scylla-image-setup.service
Swap creation on scylla-image-setup.service is too much time, it slows down startup of scylla-server.service on our IaaS image.
I think we should pre-allocate swapfile in the image build, instead of create it on startup time.
But we don't know user will run the image on which instance type, we cannot configure swapfile size from current memory size, need to hardcode the swapfile size instead.
And it should be enough size for most of instance type.

syuu1228:
Swap creation on scylla-image-setup.service is too much time, it slows down
startup of scylla-server.service on our IaaS image.
To fix the issue, pre-allocate swapfile while building the image.

Use half of diskfree to swap size, since it is maximum size allowed on
scylla_swap_setup and should be enough for any instance types.

Fixes #285

----

Please merge this only after https://github.com/scylladb/scylla/pull/9971 merged.

### DIFF
--- a/common/scylla_image_setup
+++ b/common/scylla_image_setup
@@ -42,7 +42,6 @@ if __name__ == '__main__':
             run('/opt/scylladb/scripts/scylla_coredump_setup --dump-to-raiddir', shell=True, check=True)
     else:
         run('/opt/scylladb/scripts/scylla_coredump_setup', shell=True, check=True)
-    run('/opt/scylladb/scripts/scylla_swap_setup --swap-directory /', shell=True, check=True)
     if not os.path.ismount('/var/lib/scylla'):
         print('Failed to initialize RAID volume!')
     pathlib.Path('/etc/scylla/machine_image_configured').touch()
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -52,6 +52,10 @@ def is_centos():
 def arch():
     return platform.machine()
 
+def half_of_diskfree():
+    disk = os.statvfs('/')
+    return int((disk.f_bavail * disk.f_frsize) / 2)
+
 if __name__ == '__main__':
     if os.getuid() > 0:
         print('Requires root permission.')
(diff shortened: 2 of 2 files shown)


avikivity:
I think that mkswap is only slow on a file, and not on a block device.

Perhaps we can tell cloud-init how to compute the swap size.

avikivity:
(and make it use a separate partition)

avikivity:
Here's a discussion touching it (not providing a full solution): https://stackoverflow.com/questions/60095848/how-do-you-configure-a-swap-partition-using-cloud-init

avikivity:
Perhaps with this: https://cloudinit.readthedocs.io/en/latest/topics/modules.html#scripts-per-once, and disable resize_fs. There is also https://cloudinit.readthedocs.io/en/latest/topics/modules.html#disk-setup that lets you set a partition to 50% of the disk size.

syuu1228:
@avikivity Yes, it looks like growfs at cloud-init, but something different what we expected.
Because Packer grows rootfs at **building time (packer temporary instance)**, not on Scylla AMI startup time.
At that time it's clean Ubuntu official AMI, do not contain any of our script yet, hard to configure it.
And once it grow to disk space, we cannot do online shrink.

I think we should merge this patch first, then consider adding swap partition later.

avikivity:
> @avikivity Yes, it looks like growfs at cloud-init, but something different what we expected. Because Packer grows rootfs at **building time (packer temporary instance)**, not on Scylla AMI startup time. At that time it's clean Ubuntu official AMI, do not contain any of our script yet, hard to configure it. And once it grow to disk space, we cannot do online shrink.
> 
> I think we should merge this patch first, then consider adding swap partition later.

Ok.

yaronkaikov (dismissed):

yaronkaikov (approved):

PR #296, opened 2022-02-01
fix(centos8): Centos8 is EOL, need to change mirrors

yaronkaikov:
Centos8 become EOL. need to modify mirror to keep using it for our
installations


fruch (approved):
LGTM

PR #300, opened 2022-02-14
convert scylla_cloud_io_setup to a python library

### LINKED ISSUE #298: scylla_cloud_io_setup should be implemented as library, not executable
Since we only call scylla_cloud_io_setup from `cloud_instance.io_setup()`, it should be implemented as library, not executable.
It large to merge into io_setup() method, probably we need to move scylla_cloud_io_setup to lib/io_setup.py and make it callable from io_setup().

syuu1228:
Since we only call scylla_cloud_io_setup from cloud_instance.io_setup(),
it should be implemented as library, not executable.

Fixes #298


syuu1228:
Running jenkins test at https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/next-machine-image/224

syuu1228:
Ah, looks like rebasing with 5f66a515ba48f880b9fe4fdf5fc59518d004cdd7 does not worked correctly since git thinks I removed scylla_cloud_io_setup (actually I renamed and modified).
I will rebase the file manually.

yaronkaikov:
> Got FAILURE on Jenkins test, but it looks like SCT issue, not because this PR: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/artifacts/job/artifacts-ami-test/101/console

@syuu1228 yes it;s not related. looks like you branch in pkg is not rebased ..

yaronkaikov (approved):

PR #305, opened 2022-02-20
cloudformation:Add support arm based machines

yaronkaikov:
This is a preparation for https://github.com/scylladb/scylla-pkg/issues/2616. In order to support AMI promote for arm based arch, we need to generate the cloudformation yaml file based on the relevant instance types

Adding a condition based on parameter that we will pass during the template generating process
    
Thanks to @benipeled for the providing the quick solution

### DIFF
--- a/aws/cloudformation/scylla.yaml.j2
+++ b/aws/cloudformation/scylla.yaml.j2
@@ -75,6 +75,7 @@ Parameters:
       - 'true'
     Default: 'false'
 
+{% if 'x86_64' in arch %}
   InstanceType:
     Type: String
     Default: i3.large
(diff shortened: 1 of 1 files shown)


fruch:
Should you also revert #302  in this one ?

yaronkaikov:
@fruch Since it's not in master (next-machine-image is broken)  i reverted this commit

yaronkaikov:
In any case. don't merge, since i will verify it with jenkins job (locally it's working)

yaronkaikov:
Verification:
https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/1251/

@fruch Can you please approve and merge

aws/cloudformation/scylla.yaml.j2
```diff
@@ -94,6 +95,21 @@ Parameters:
       - i3en.12xlarge
       - i3en.24xlarge
       - i3en.metal
+{% endif %}
+{% if 'aarch64' in arch %}
```
benipeled, line 99: You can short it with `{% elif 'aarch64' in arch %}`

benipeled (changes requested):

benipeled (approved):

bentsi (approved):

PR #310, opened 2022-02-21
Drop all dependencies to scylla_util library

syuu1228:
We recently moved all cloud related code from scylla repository to
scylla-machine-image, to remove cross-dependencies between two repositories.
However, we still have cross-dependencies problem because moved code
uses scylla_util library which is on scylla repository.
This is still not easy to maintain, and also difficult to implement
testcase because the code won't work without scylla repository.

To resolve this, drop all dependencies to scylla_util library.

### DIFF
--- a/common/scylla_ec2_check
+++ b/common/scylla_ec2_check
@@ -19,9 +19,7 @@ import os
 import sys
 import re
 import argparse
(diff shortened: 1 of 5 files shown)


syuu1228:
Merged to #312, since it requires for aws_instance testcase

PR #324, opened 2022-03-03
gce:Build image based on Centos-Stream-8 instead of centos8 

yaronkaikov:
[](https://github.com/yaronkaikov)Since centos8 is EOL, we need to switch to centos-stream-8 to continue
supporting gce images (which are rpm based).


yaronkaikov:
Verified:
https://jenkins.scylladb.com/job/scylla-4.4/job/gce-image/142/

fruch (commented):
LGTM

PR #334, opened 2022-04-11
ubuntu: drop unneeded packages

syuu1228:
This remove unneeded packages on Ubuntu image, includes:
- Drop snapd and modemmanager
- Purge uninstalled packages
- Run autoremove after uninstalling packages

PR #336, opened 2022-04-21
scylla_install_image:add ssm-agent installation for arm

yaronkaikov:
In https://github.com/scylladb/scylla-machine-image/pull/334/commits/434795fcd0522457d843dfb695f46493330486d0
snap installation was removed and we started to use ssm-agent deb package. Since we are building AMI based on ARM as well. we need to support both arches

### DIFF
--- a/dist/debian/build_deb.sh
+++ b/dist/debian/build_deb.sh
@@ -81,6 +81,7 @@ pkg_install python3
 pkg_install python3-devel
 pkg_install python3-pip
 
+git config --global --add safe.directory "$BUILDDIR"/scylla-machine-image
 echo "Building in $PWD..."
 
 VERSION=$(./SCYLLA-VERSION-GEN)
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -138,7 +138,13 @@ if __name__ == '__main__':
             pyver = '3'
 
             # install .deb version of ssm-agent since we dropped snapd version
-            run('curl -L -o /tmp/amazon-ssm-agent.deb https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/debian_amd64/amazon-ssm-agent.deb')
+            match arch():
+                case "x86_64":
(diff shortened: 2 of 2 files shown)


yaronkaikov:
Verified with: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/1325/

dist/debian/build_deb.sh
```diff
@@ -81,6 +81,7 @@ pkg_install python3
 pkg_install python3-devel
 pkg_install python3-pip
 
+git config --global --add safe.directory "$BUILDDIR"/scylla-machine-image
```
bentsi, line 84: why did you add this?

yaronkaikov: https://github.com/scylladb/scylla-machine-image/runs/6108039418?check_suite_focus=true 

because of that...
Added a Ref in commit description

packer/scylla_install_image
```diff
@@ -138,7 +138,10 @@ if __name__ == '__main__':
             pyver = '3'
 
             # install .deb version of ssm-agent since we dropped snapd version
-            run('curl -L -o /tmp/amazon-ssm-agent.deb https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/debian_amd64/amazon-ssm-agent.deb')
+            if platform.machine() == 'x86_64':
```
bentsi, line 141:
```suggestion
    arch = platform.machine()
    match arch:
        case "x86_64":
            run('curl -L -o /tmp/amazon-ssm-agent.deb https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/debian_amd64/amazon-ssm-agent.deb')
        case "aarch64":
            run('curl -L -o /tmp/amazon-ssm-agent.deb https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/debian_arm64/amazon-ssm-agent.deb'
        case _:
            raise Exception("Unknown arch")

```

bentsi (approved):

yaronkaikov:
Dropped from next due some syntax error in code. it was merged while verification still was in progress

PR #337, opened 2022-04-21
scylla_install_image:add ssm-agent installation for arm

yaronkaikov:
In 434795f snap installation was removed and we started to use ssm-agent deb
package. Since we are building AMI based on ARM as well. we need to
support both arch


yaronkaikov:
Verification: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/1327/

bentsi (approved):

PR #338, opened 2022-04-21
build.yml: workaround git permission bug for checkout dir

yaronkaikov:
It seems that a new permission bug was intreduced from git version 2.35.1
Until we have a proper fix, adding this workaround (Ref: https://github.com/actions/checkout/issues/760)


yaronkaikov:
Closing. open for debug

PR #340, opened 2022-04-26
Add scylla-api-client to scylla-image

benipeled:
Following @bhalevy request in scylla-pkg/issues/2579, this commit adds
the scylla-api-client tool to scylla-image on aws, gce and azure.

@syuu1228, the scylla_install_image use easy_install for python packages installation and there is a comment 
```
# use easy_install instead of pip, because easy_install requires
# fewer dependencies, no need to install gcc, it makes our AMI smaller.
```
Can you please elaborate on these image-size differences? is it worth dealing with easy_install which is deprecated and should be removed one day?

Closes https://github.com/scylladb/scylla-pkg/issues/2579

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -46,6 +46,38 @@ def half_of_diskfree():
     disk = os.statvfs('/')
     return int((disk.f_bavail * disk.f_frsize) / 2)
 
+
+def install_required_packages(distro):
+    """
+    The python packages will be installed with easy_install instead of pip, because
+    easy_install requires fewer dependencies, no need to install gcc, it makes our AMI smaller.
+    """
+
+    deb_packages = ['python3-pystache', 'python3-lockfile', 'python3-setuptools']
+    rpm_packages = ['pystache', 'python-daemon', 'python-setuptools']
+    pip_packages = ['scylla-api-client']
+
+    if distro == 'centos':
+        run('yum install -y epel-release')
+
+        for package in rpm_packages:
+            run(f'yum install -y {package}')
+
+        for package in pip_packages:
+            run(f'python -m easy_install {package}')
+
+    elif distro == 'ubuntu':
+        for package in deb_packages:
+            run(f'apt-get install -y {package}')
+
+        for package in pip_packages:
+            run(f'python3 -m easy_install {package}')
+
+    else:
+        print(f"Unknown distro {distro}")
+        sys.exit(1)
+
+
 if __name__ == '__main__':
     if os.getuid() > 0:
         print('Requires root permission.')
@@ -91,6 +123,7 @@ if __name__ == '__main__':
             os.remove('/etc/yum.repos.d/scylla_install.repo')
         if args.repo_for_update:
             run('curl -L -o /etc/yum.repos.d/scylla.repo {REPO_FOR_UPDATE}'.format(REPO_FOR_UPDATE=args.repo_for_update))
+        install_required_packages(distro)
     else:
         run('apt-key adv --keyserver keyserver.ubuntu.com --recv-keys d0a112e067426ab2')
         if args.repo_for_install:
@@ -118,13 +151,10 @@ if __name__ == '__main__':
         run('apt-get purge -y modemmanager')
         # drop packages does not needed anymore
         run('apt-get autoremove --purge -y')
+        install_required_packages(distro)
 
     if args.target_cloud == 'aws':
-        # use easy_install instead of pip, because easy_install requires
-        # fewer dependencies, no need to install gcc, it makes our AMI smaller.
         if distro == 'centos':
-            run('yum install -y epel-release')
-            run('yum install -y pystache python-daemon python-setuptools')
             # Default install prefix on CentOS7 is /usr not /usr/local, but we need
             # to align path for executable, so specify script-dir to /usr/local/bin
             run('python -m easy_install --script-dir /usr/local/bin https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-latest.tar.gz')
@@ -133,7 +163,6 @@ if __name__ == '__main__':
             # XXX: we want install dependencies from distro repo, but
             # python-daemon-2.2 on Ubuntu 20.04 is too new for aws-cfn-bootstrap,
             # so install python-daemon-2.1 from pypi
-            run('apt-get install -y python3-pystache python3-lockfile python3-setuptools')
             run('python3 -m easy_install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz')
             pyver = '3'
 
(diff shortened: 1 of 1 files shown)


fruch:
> Following @bhalevy request in scylla-pkg/issues/2579, this commit adds the scylla-api-client tool to scylla-image on aws, gce and azure.
> 
> @syuu1228, the scylla_install_image use easy_install for python packages installation and there is a comment
> 
> ```
> # use easy_install instead of pip, because easy_install requires
> # fewer dependencies, no need to install gcc, it makes our AMI smaller.
> ```
> 
> Can you please elaborate on these image-size differences? is it worth dealing with easy_install which is deprecated and should be removed one day?

see
https://packages.ubuntu.com/impish/python3-pip

`build-essential` is only recommendation, pip isn't depended on GCC. (some packages are, which mean they won't be installable also with easy_install)

benipeled:
> `build-essential` is only recommendation, pip isn't depended on GCC. (some packages are, which mean they won't be installable also with easy_install)

Indeed,
Since the comment (and the implementation) was added by @syuu1228 I want him to take a look at it before we change it to `pip` in case there is some reason for using easy_install we should take into account.

syuu1228 (approved):
@benipeled Looks okay if we have to use PIP package for scylla-api-client, but how about to add rpm/deb packaging for scylla-api-client?
I can provide a patch for it.
But maybe we can start this with PIP package now, and add rpm/deb packaging later.

benipeled:
The easy_install vs pip question is not specific about scylla-api-client but also for aws-cfn-bootstrap installed by easy_install and if we have no reason for that I prefer to change it to pip,

scylla-api-client is a python package and (at least for now) I see no good reason for dealing with deb/rpm packages

benipeled:
> aws-cfn-bootstrap is a 3rdparty tools, and they don't provide .deb package so there is no choice, we have to use easy_install, so we can't discuss with this. I hope they will also provide .deb package but they only have rpm and PIP today.
> 
Sure, and because aws-cfn-bootstrap is built as a python package we can use `pip install <url-to-the-zip-file>` or even directly to PyPI by `pip install aws-cfn-bootstrap` instead of easy_install which is deprecated,

> On Scylla packages, we chosen to use distribution default packaging system, since it's not just python but also native binaries, such as scylla-server itself. Even we use multiple languages to develop Scylla, we decided to use single packaging system, users can install or update our packages by simply distribution standard way, such as `yum install` / `yum update` or `apt install` / `apt upgrade`.
> 
I can see your point and it makes sense,
The question in this case, is whether scylla-api-client is considered as a tool that is part of the Scylla product or as a helper tool that is distributed separately (it is distributed separately),

As you suggested, let's start with the pip package and we can add a deb package later (on [scylla-api-client](https://github.com/scylladb/scylla-api-client) project)

bhalevy:
> As you suggested, let's start with the pip package and we can add a deb package later (on [scylla-api-client](https://github.com/scylladb/scylla-api-client) project)

pip could also be used at packaging time, no?  as part of `ninja release-dist`, so to pack scylla-api-client even into the relocatable package and then respectively to rpm/dev, and finally to the machine image.  I'm not an expert on the distribution process so maybe I'm talking nonsense...

benipeled:
> pip could also be used at packaging time, no? as part of `ninja release-dist`, so to pack scylla-api-client even into the relocatable package and then respectively to rpm/dev, and finally to the machine image. I'm not an expert on the distribution process so maybe I'm talking nonsense...

Thecnicly it's possible but then scylla-api-client release will stick to scylla release and in case of an important fix to scylla-api-client it will not be possible to update scylla-api-client only but it will require waiting for Scylla release,
(or using a pre-release relocatable but it's out of the release flow),

If no other concerns, I'll run 2nd verification, and let's merge it,

/cc @avikivity

packer/scylla_install_image
```diff
@@ -46,6 +46,38 @@ def half_of_diskfree():
...
+
+    if distro == 'centos':
+        run('yum install -y epel-release')
+
+        for package in rpm_packages:
+            run(f'yum install -y {package}')
+
+        for package in pip_packages:
+            run(f'python -m easy_install {package}')
+
```
avikivity, line 68: Why are we installing some packages via rpm and some packages via pip? It's a recipe for confusion and non-upgraded packages.

benipeled: As descussed  above
- aws-cfn-bootstrap - no deb package so we use the provided tar with easy_install / pip
- scylla-api-client - a python project built as a python package. we might later add/switch to a deb package

syuu1228:
> pip could also be used at packaging time, no? as part of `ninja release-dist`, so to pack scylla-api-client even into the relocatable package and then respectively to rpm/dev, and finally to the machine image. I'm not an expert on the distribution process so maybe I'm talking nonsense...

Ah, if we are okay to install scylla-api-client unconditionally as a part of scylla-python3 rpm/deb package (instead of optional package), your suggestion is better and much easier.
(We are currently downloading some python modules from PIP and making scylla-python3 rpm/deb package with these modules)

syuu1228:
> Thecnicly it's possible but then scylla-api-client release will stick to scylla release and in case of an important fix to scylla-api-client it will not be possible to update scylla-api-client only but it will require waiting for Scylla release, (or using a pre-release relocatable but it's out of the release flow),

Ah right, it will be part of scylla releases if we contain it to scylla-python3 package. If we want to have different release flow we shouldn't do that.

bhalevy:
> Ah, if we are okay to install scylla-api-client unconditionally as a part of scylla-python3 rpm/deb package (instead of optional package), your suggestion is better and much easier.

I see no problem with installing it unconditionally.

> > Thecnicly it's possible but then scylla-api-client release will stick to scylla release and in case of an important fix to scylla-api-client it will not be possible to update scylla-api-client only but it will require waiting for Scylla release, (or using a pre-release relocatable but it's out of the release flow),
> 
> Ah right, it will be part of scylla releases if we contain it to scylla-python3 package. If we want to have different release flow we shouldn't do that.

scylla-api-client currently has no dependencies on a particular scylla release, by design, so it can be considered orthogonal to the scylla installation and theoretically it could be updated independently.
But practically, I don't think this provides much benefits.
If we bundle it with the scylla release we can make sure it's tested with that release, otherwise we'll need a much wider testing matrix of scylla-api-client versions vs. scylla versions.

benipeled:
@bhalevy @syuu1228 following our discussion,

1. The scylla-api-client will be added to the scylla packaging flow to ensure it will be upgraded with other scylla-packages on scylla-images.
2. The pip option for scylla-api-client will remain to allow self-installation regardless of scylla-server.

@syuu1228 please take care for 1

bhalevy:
> @bhalevy @syuu1228 following our discussion,
> 
> 1. The scylla-api-client will be added to the scylla packaging flow to ensure it will be upgraded with other scylla-packages on scylla-images.
> 2. The pip option for scylla-api-client will remain to allow self-installation regardless of scylla-server.
> 
> @syuu1228 please take care for 1

Thanks!

syuu1228:
Posted PRs to add support scylla-api-client.

Issue #341, opened 2022-04-27
XFS online discard: fstrim.timer service is enabled after the start of a Scylla instance from a machine image


enaydanov:
We added `discard` option to the XFS mount recently (https://github.com/scylladb/scylla/commit/a19d00ef9b8b50ce520bb2cfbf5bc00b30de5cda) and fstrim is pointless when online discard is enabled (according to Avi.)

But for instances which run from Scylla machine images fstrim.timer service is still enabled.

fruch:
@enaydanov can you share the logs ? and which images did you used ?

fruch:
@enaydanov FYI, fstrim is enabled by default on ubuntu, so fixing it here won't be enough. we should consider disabling it as part of scylla_setup

fruch:
this is the commandline, and there's no filter
```
ExecStart=/sbin/fstrim --listed-in /etc/fstab:/proc/self/mountinfo --verbose --quiet-unsupported
```

```
❯ /sbin/fstrim --help

Usage:
 fstrim [options] <mount point>

Discard unused blocks on a mounted filesystem.

Options:
 -a, --all                trim mounted filesystems
 -A, --fstab              trim filesystems from /etc/fstab
 -I, --listed-in <list>   trim filesystems listed in specified files
 -o, --offset <num>       the offset in bytes to start discarding from
 -l, --length <num>       the number of bytes to discard
 -m, --minimum <num>      the minimum extent length to discard
 -v, --verbose            print number of discarded bytes
     --quiet-unsupported  suppress error messages if trim unsupported
 -n, --dry-run            does everything, but trim

 -h, --help          display this help
 -V, --version       display version

Arguments:
 <num> arguments may be followed by the suffixes for
   GiB, TiB, PiB, EiB, ZiB, and YiB (the "iB" is optional)

For more details see fstrim(8).
```

regardless on our own images it would be safe to disable, wouldn't it ?

slivne:
Yes on our images we should disable

PR #343, opened 2022-04-28
XFS online discard: Disable fstrim.timer if active

fruch:
Some distro has fstrim enabled by default, since we
are using XFS with online discard, we don't need fstrim
running

Fixes: #341

### DIFF
--- a/common/scylla_image_setup
+++ b/common/scylla_image_setup
@@ -30,6 +30,9 @@ if __name__ == '__main__':
             run('systemctl daemon-reload', shell=True, check=True)
             run('systemctl enable var-lib-systemd-coredump.mount', shell=True, check=True)
             run('systemctl start var-lib-systemd-coredump.mount', shell=True, check=True)
+            # some distro has fstrim enabled by default, since we are using XFS with online discard, we don't need fstrim
+            run('systemctl is-active -q fstrim.timer && systemctl disable fstrim.timer', shell=True, check=True)
+
     if not os.path.ismount('/var/lib/scylla'):
         print('Failed to initialize RAID volume!')
     pathlib.Path('/etc/scylla/machine_image_configured').touch()


common/scylla_image_setup
```diff
@@ -30,6 +30,9 @@ if __name__ == '__main__':
             run('systemctl daemon-reload', shell=True, check=True)
             run('systemctl enable var-lib-systemd-coredump.mount', shell=True, check=True)
             run('systemctl start var-lib-systemd-coredump.mount', shell=True, check=True)
+            # some distro has fstrim enabled by default, since we are using XFS with online discard, we don't need fstrim
+            run('if [[ `systemctl is-active fstrim.timer` == "active" ]]; then systemctl disable fstrim.timer; fi', shell=True, check=True)
```
enaydanov, line 34: `is-active` command has a return code. So, you can just do `systemctl is-active -q fstrim.timer && ...`

fruch:
@yaronkaikov FYI, see the commit here that fixes the PR build for DEB

common/scylla_image_setup
fruch: your'e right, didn't notice the `-q`, when called without it the return code is 0 either way.

fruch: fixed

dist/debian/build_deb.sh
```diff
@@ -98,6 +98,8 @@ fi
 rm -rf "$BUILDDIR"
 mkdir -p "$BUILDDIR"/scylla-machine-image
 
+git config --global --add safe.directory $(realpath $(dirname $0)"/../..")
```
yaronkaikov, line 101: AFAIK this will break packing building using `dbuild` please verify it before

yaronkaikov (changes requested):

enaydanov (dismissed):

dist/debian/build_deb.sh
avikivity: It's preposterous to insert global configuration commands into a build script. A build script should only touch its build directly, nothing else.

avikivity: Also, what does git configuration have to do with disabling fstrim? Please practice patch hygiene.

fruch: removed this commit, and I'll try fixing it in a different way

yaronkaikov (approved):

enaydanov:
@fruch can you backport it to 5.0 and 2022.1 ?

fruch:
> @fruch can you backport it to 5.0 and 2022.1 ?

backported to boths next branches

PR #345, opened 2022-05-01
build_deb_image.sh:remove default tag for branch name

yaronkaikov:
When QA running jobs which are relaying on recent AMI, the first run `hydra list-ami-branch -r eu-west-1 master:all`

The query looks for AMI with tag name `BRANCH=master`. when building local AMI for debug if no `--branch <name>` is provided we have a default of `master` , removing the default so QA will get the only relevant images

Closes: https://github.com/scylladb/scylla-pkg/issues/2856

### DIFF
--- a/packer/build_deb_image.sh
+++ b/packer/build_deb_image.sh
@@ -8,7 +8,6 @@ REALDIR=$(dirname $(readlink -f "$0"))
 source "$REALDIR"/../SCYLLA-VERSION-GEN
 
 BUILD_ID=$(date -u '+%FT%H-%M-%S')
-BRANCH="master"
 OPERATING_SYSTEM="ubuntu20.04"
 DIR=$(dirname $(realpath -se $0))
 PDIRNAME=$(basename $(realpath -se $DIR/..))


fruch (approved):
LGTM

fruch:
@yaronkaikov did we clear the tags from older local built AMIs ?

yaronkaikov:
> @yaronkaikov did we clear the tags from older local built AMIs ?

no. i will do it later

fgelcer:
IMHO it would be even better to have something like `BRANCH="dev"` instead of leaving it empty, so we will know for sure that, if nothing specified, it will go to some `dev` branch name, instead of any of existing one...

fgelcer:
> IMHO it would be even better to have something like `BRANCH="dev"` instead of leaving it empty, so we will know for sure that, if nothing specified, it will go to some `dev` branch name, instead of any of existing one...

@yaronkaikov @fruch ^^

yaronkaikov:
Verification: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/1338/

yaronkaikov:
> > IMHO it would be even better to have something like `BRANCH="dev"` instead of leaving it empty, so we will know for sure that, if nothing specified, it will go to some `dev` branch name, instead of any of existing one...
> 
> @yaronkaikov @fruch ^^

@fgelcer i will do a follow up change in the images naming convention so it will make more sense. one of the things is to set the branch name in case someone is creating a debug image with the username inside the branch so it will be easy to identify 

I

yaronkaikov:
@fgelcer https://github.com/scylladb/scylla-machine-image/issues/346

fgelcer:
> > > IMHO it would be even better to have something like `BRANCH="dev"` instead of leaving it empty, so we will know for sure that, if nothing specified, it will go to some `dev` branch name, instead of any of existing one...
> > 
> > 
> > @yaronkaikov @fruch ^^
> 
> @fgelcer i will do a follow up change in the images naming convention so it will make more sense. one of the things is to set the branch name in case someone is creating a debug image with the username inside the branch so it will be easy to identify
> 
> I

sounds great to me, thank you.

PR #348, opened 2022-05-02
print subprocess error message when capture_output=True

syuu1228:
We currently does not able to get any error message from subprocess when we specified capture_output=True on subprocess.run().
This is because CalledProcessError does not print stdout/stderr when it raised, and we don't catch the exception, we just let python to cause Traceback.
Result of that, we only able to know exit status and failed command but
not able to get stdout/stderr.

To resolve this, add wrapper function "out()" for capture output, and
print stdout/stderr with error message inside the function.

This is scylla-machine-image version of https://github.com/scylladb/scylla/pull/10391

### DIFF
--- a/lib/scylla_cloud.py
+++ b/lib/scylla_cloud.py
@@ -18,9 +18,27 @@ import glob
 import distro
 import base64
 import datetime
-from subprocess import run, DEVNULL
+from subprocess import run, CalledProcessError
 from abc import ABCMeta, abstractmethod
 from lib.scylla_cloud_io_setup import aws_io_setup, gcp_io_setup, azure_io_setup
+
+
+def out(cmd, shell=True, timeout=None, encoding='utf-8', ignore_error=False, user=None, group=None):
+    try:
+        res = run(cmd, capture_output=True, shell=shell, timeout=timeout, check=not ignore_error, encoding=encoding, user=user, group=group)
+    except CalledProcessError as e:
+        print(f'''
+Command '{cmd}' returned non-zero exit status: {e.returncode}
+----------  stdout  ----------
+{e.stdout.strip()}
+------------------------------
+----------  stderr  ----------
+{e.stderr.strip()}
+------------------------------
+'''[1:-1])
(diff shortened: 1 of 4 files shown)


lib/scylla_cloud.py
```diff
@@ -21,6 +21,21 @@
 from abc import ABCMeta, abstractmethod
 from lib.scylla_cloud_io_setup import aws_io_setup, gcp_io_setup, azure_io_setup
 
+
+def out(cmd, shell=True, timeout=None, encoding='utf-8', ignore_error=False):
+    res = run(cmd, capture_output=True, shell=shell, timeout=timeout, check=False, encoding=encoding)
+    if not ignore_error and res.returncode != 0:
+        print(f'Command \'{cmd}\' returned non-zero exit status: {res.returncode}')
```
benipeled, line 28: Not critical but maybe here we can catch the `CalledProcessError` exception and use the `CalledProcessError.output`

lib/scylla_cloud.py
```diff
@@ -21,6 +21,21 @@
...
+        print('----------  stdout  ----------')
+        print(res.stdout, end='')
+        print('------------------------------')
+        print('----------  stderr  ----------')
+        print(res.stderr, end='')
+        print('------------------------------')
```
benipeled, line 34: You can simplfy it with
```python
print('''f
----------  stdout  ----------
{res.stdout}
.
.
.

''')

lib/scylla_cloud.py
syuu1228: switched to catching CalledProcessError exception.

lib/scylla_cloud.py
syuu1228: Switched to here-document style.

syuu1228:
Rebased with latest next, and fixed code style things pointed out by @benipeled.

syuu1228:
Rebased with next

syuu1228:
@yaronkaikov ping

PR #359, opened 2022-05-24
set pre-release to be semantic versions

yaronkaikov:
Today we are not using semantic versioning for our pre-release, Let's switch to the following convertion for both rpm and deb packages:
```
4.5.0~rc0 .. 4.5.0~rc1 .. 4.5.0
```

For that we need to remove the replacing of `.rc` with `~rc`

Ref: https://github.com/scylladb/scylla/issues/9543


fruch:
@yaronkaikov how this is changing the docker images ? i.e. the original bug talks about using scylla in operator.

dist/debian/build_deb.sh
```diff
@@ -90,11 +90,6 @@ PRODUCT=$(cat build/SCYLLA-PRODUCT-FILE)
 BUILDDIR=build/debian
 PACKAGE_NAME="$PRODUCT-machine-image"
 
-if echo $SCYLLA_VERSION | grep rc >/dev/null ; then
```
fruch, line 93: shouldn't this be fixed at the source, i.e. in scylla repo ? or i'm missing something

fruch: i.e. scylla itself saying the wrong version:
```
❯ /home/fruch/.ccm/scylla-repository/release/5.0.rc3/scylla/bin/scylla --version
5.0.rc3-0.20220406.f92622e0d
```

yaronkaikov: The fix is only relevant from next time we will branch. following this issue (which was originally for docker images), we decided to align the pre-release so it will match both rpm and deb packages

fruch (approved):
LGTM

PR #360, opened 2022-05-25
azure: fix is recommended instance method

soyacz:
method `self.is_unsupported_instance_class()` always was returning `False` making `is_recommended_instance` always return `False` causing error in io_setup


yaronkaikov (approved):
LGTM
/cc @syuu1228

PR #362, opened 2022-05-30
AWS: Support IMDSv2 on aws_instance

syuu1228:
Currently, our AMI fails to start scylla-image-setup when the instance
configured as "Metadata version: V2 only".
To fix this we need to implement IMDSv2 support on aws_instance class.

Fixes scylladb/scylla#9987

### DIFF
--- a/lib/scylla_cloud.py
+++ b/lib/scylla_cloud.py
@@ -17,16 +17,17 @@ import socket
 import glob
 import distro
 import base64
+import datetime
 from subprocess import run, DEVNULL
 from abc import ABCMeta, abstractmethod
 from lib.scylla_cloud_io_setup import aws_io_setup, gcp_io_setup, azure_io_setup
 
 # @param headers dict of k:v
-def curl(url, headers=None, byte=False, timeout=3, max_retries=5, retry_interval=5):
+def curl(url, headers=None, method=None, byte=False, timeout=3, max_retries=5, retry_interval=5):
     retries = 0
     while True:
         try:
-            req = urllib.request.Request(url, headers=headers or {})
+            req = urllib.request.Request(url, headers=headers or {}, method=method)
             with urllib.request.urlopen(req, timeout=timeout) as res:
                 if byte:
                     return res.read()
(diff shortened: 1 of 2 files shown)


syuu1228:
@yaronkaikov Ah, it is because test_aws_instance.py has metadata server mock, and it only supports IMDSv1 so all metadata access test failing.
I will fix that.

syuu1228:
Test fixed.

yaronkaikov:
@syuu1228 How can we verify it's working? is there a special instance type?
in any case please run https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/next-machine-image/ based on your branch and let's make sure it is passing. Also please run https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/ so can verify we didn't break `arm` as well

syuu1228:
> @syuu1228 How can we verify it's working? is there a special instance type?

There is an option to disable IMDSv1 and force use IMDSv2 (force use http token).
In Web UI it's "Metadata version: V2 only" in "Advanced details", in CLI it's like this:
```
aws ec2 run-instances 
    --image-id ami-0abcdef1234567890 
    --instance-type c3.large 
	...
    --metadata-options "HttpEndpoint=enabled,HttpTokens=required"
```
see: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html

I guess we can do same in our tests on Jenkins.

syuu1228:
Tests successed.

syuu1228:
> > @syuu1228 How can we verify it's working? is there a special instance type?
> 
> 
> 
> There is an option to disable IMDSv1 and force use IMDSv2 (force use http token).
> 
> In Web UI it's "Metadata version: V2 only" in "Advanced details", in CLI it's like this:
> 
> ```
> 
> aws ec2 run-instances 
> 
>     --image-id ami-0abcdef1234567890 
> 
>     --instance-type c3.large 
> 
> 	...
> 
>     --metadata-options "HttpEndpoint=enabled,HttpTokens=required"
> 
> ```
> 
> see: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html
> 
> 
> 
> I guess we can do same in our tests on Jenkins.

I fogot describing, I manually tested the patch with "V2 only" mode and it worked.

yaronkaikov (approved):

Issue #357, opened 2022-05-20
support cloud-init along with Scylla User Data


soyacz:
Currently scylla-mechine-image configure script takes over user-data and does not allow to use cloud-init compatible input.
This prevents from using feature rich cloud-init and replaces it with something limited only to scylla.
E.g. in SCT we would like to use cloud-init with .yaml formatted files (see https://cloudinit.readthedocs.io/en/latest/topics/examples.html)
to configure instances.
Also our customers might want to use cloud-init and not resign of using scylla-machine-image configuration.

Request is about adding support for cloud-init compatible scripts.
I have ideas on how to do it:
1. Create cloud-init module.
Will require module implementation and installation in proper place. Docs about creating modules: https://cloudinit.readthedocs.io/en/latest/topics/module_creation.html
2. As currently it is done but add yaml support to user-data and parse it.
3. Create separate Scylla configuration file
`cloud-init` .yaml files can define `write_files` section which allows users to create files at any place in the system. Idea is to use it to create scylla configuration json and use it instead of scylla specific user-data. This should be very easy and we could maintain backward compatibility (e.g. if we find `#cloud-config` preamble in user-data, use file, or just use file and if not found, fallback to current solution)

soyacz:
@bentsi ^^

fruch:
FYI, @yaronkaikov @syuu1228

@tzach, did we ever got requests/questions like this from customers ?

soyacz:
@tzach @slivne If you agree we want this I'd be glad to implement it (option 3 - with backward compatibility).

PR #363, opened 2022-06-01
Add yaml and mime multipart support for user-data

fruch:
Now we can support configuring the user-data with yaml
and with mime multipart

mime multipart is important for enabling a user of our
images to configure both scylla and all other cloud-init
capabilities and configurations

Closes: #357


common/scylla_configure.py
```diff
@@ -85,10 +84,24 @@ def save_scylla_yaml(self):
...
+                # scylla-machine-image configuration out of it
+                message = message_from_string(raw_user_data)
+                if message.is_multipart():
+                    for i in message.walk():
+                        if i.get_content_type() in ['x-scylla/json', 'x-scylla/yaml']:
+                            raw_user_data = i.get_payload()
+
+                # try parse yaml, and and fallback to parsing json
+                self._instance_user_data = {}
+                if not self._instance_user_data and raw_user_data:
```
soyacz, line 100: Here `self._instance_user_data` will always be `False` - no need to verify it.

soyacz (approved):
LGTM
Only concern to docs: Shall we still call it user_data_v2? Shouldn't we differentiate from previous version by e.g. making it `2.1`? Somewhere we need to mark which versions of Scylla support it.

benipeled (approved):

yaronkaikov:
@fruch How did you verify it;s working/not breaking anything ?

fruch:
> @fruch How did you verify it;s working/not breaking anything ?

for now, only the unit-tests.

fruch:
> LGTM Only concern to docs: Shall we still call it user_data_v2? Shouldn't we differentiate from previous version by e.g. making it `2.1`? Somewhere we need to mark which versions of Scylla support it.

I'll change the version to 3, so we could identify it in SCT (i.e. SCT is the old place someone cares about this version number)

common/scylla_configure.py
fruch: fixed

fruch:
> > @fruch How did you verify it;s working/not breaking anything ?
> 
> for now, only the unit-tests.


It's a very localized change, and the unit-tests are covering the original json based user-data, and empty cases.
this change doesn't touch anything else (just documentation, and update `user_data_format_version`)

common/scylla_configure.py
```diff
@@ -85,10 +84,24 @@ def save_scylla_yaml(self):
...
+
+                # Try reading mime multipart message, and extract
+                # scylla-machine-image configuration out of it
+                message = message_from_string(raw_user_data)
+                if message.is_multipart():
+                    for i in message.walk():
+                        if i.get_content_type() in ['x-scylla/json', 'x-scylla/yaml']:
+                            raw_user_data = i.get_payload()
+
+                # try parse yaml, and and fallback to parsing json
```
vponomaryov, line 98: Typo: doeble `and`

common/scylla_configure.py
```diff
@@ -85,10 +84,24 @@ def save_scylla_yaml(self):
...
+                self._instance_user_data = {}
+                if raw_user_data:
+                    self._instance_user_data = yaml.safe_load(raw_user_data)
```
vponomaryov, line 101: If we have raw_user_data and it is json then we do not reach line 103 and `self._instance_user_data` becomes string-with-json-data instead of expected python dict.
Or it is expected situation?

README.md
```diff
@@ -65,6 +66,50 @@ Spinning a new node connecting to “10.0.219.209” as a seed, and installing c
...
+  experimental: true
+  seed_provider:
+    - class_name: org.apache.cassandra.locator.SimpleSeedProvider
+      parameters:
+        - seeds: 10.0.219.209
+  post_configuration_script: "#! /bin/bash\nyum install cloud-init-cfn"
+  start_scylla_on_first_boot: true
+```
+
+#### using mine multipart
```
vponomaryov, line 82: s/mine/mime/

README.md
```diff
@@ -65,6 +66,50 @@ Spinning a new node connecting to “10.0.219.209” as a seed, and installing c
...
+
+If other feature of cloud-init are needed, one can use mine multipart, and pass
```
vponomaryov, line 84: s/mine/mime/

README.md
```diff
@@ -65,6 +66,50 @@ Spinning a new node connecting to “10.0.219.209” as a seed, and installing c
...
+a json/yaml with `x-scylla/yaml` or `x-scylla/json`
+
+more information on cloud-init multipart user-data:
+
+https://cloudinit.readthedocs.io/en/latest/topics/format.html#mime-multi-part-archive
+
+```mine
```
vponomaryov, line 91: s/mine/mime/

source/user_data_v3.rst
```diff
@@ -89,3 +91,48 @@ Spinning a new node connecting to "10.0.219.209" as a seed, and installing cloud
...
+      cluster_name: test-cluster
+      experimental: true
+      seed_provider:
+        - class_name: org.apache.cassandra.locator.SimpleSeedProvider
+          parameters:
+            - seeds: 10.0.219.209
+      post_configuration_script: "#! /bin/bash\nyum install cloud-init-cfn"
+      start_scylla_on_first_boot: true
+
+using mine multipart
```
vponomaryov, line 108: s/mine/mime/

source/user_data_v3.rst
```diff
@@ -89,3 +91,48 @@ Spinning a new node connecting to "10.0.219.209" as a seed, and installing cloud
...
+++++++++++++++++++++
+
+If other feature of cloud-init are needed, one can use mine multipart, and pass
```
vponomaryov, line 111: s/mine/mime/

tests/test_scylla_configure.py
```diff
@@ -162,6 +178,33 @@ def test_postconfig_script_with_bad_exit_code(self):
...
+                scylla_yaml=dict(
+                    cluster_name=self.test_cluster_name,
+                    listen_address=ip_to_set,
+                    broadcast_rpc_address=ip_to_set,
+                    seed_provider=[{
+                        "class_name": "org.apache.cassandra.locator.SimpleSeedProvider",
+                        "parameters": [{"seeds": ip_to_set}]}],
+                )
+            )
+        for data_type in ['json', 'yaml']:
```
vponomaryov, line 194: Same here. Please use tuple instead of lists for collections which are not expected to be changed as mentioned above.

common/scylla_configure.py
```diff
@@ -85,10 +84,24 @@ def save_scylla_yaml(self):
...
+                if not self._instance_user_data and raw_user_data:
+                    self._instance_user_data = json.loads(raw_user_data)
+                LOGGER.debug("JSON/YAML parsed user-data: %s", self._instance_user_data)
```
vponomaryov, line 104: I guess would be better to log messages when we know for sure the type.

vponomaryov (changes requested):

common/scylla_configure.py
```diff
@@ -85,10 +84,24 @@ def save_scylla_yaml(self):
...
+                raw_user_data = self.cloud_instance.user_data.strip()
                 LOGGER.info("Got user-data: %s", raw_user_data)
-                self._instance_user_data = json.loads(raw_user_data) if raw_user_data.strip() else {}
-                LOGGER.debug("JSON parsed user-data: %s", self._instance_user_data)
+
+                # Try reading mime multipart message, and extract
+                # scylla-machine-image configuration out of it
+                message = message_from_string(raw_user_data)
+                if message.is_multipart():
+                    for i in message.walk():
```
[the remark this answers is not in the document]
fruch: it was lazy me, would fix

common/scylla_configure.py
fruch: it's not important, at this stage it's already python dict. i'll remove it from the debug print

common/scylla_configure.py
KnifeyMoloko, line 101:
```suggestion
                    self._instance_user_data = yaml.safe_load(raw_user_data) or json.loads(raw_user_data)
```

KnifeyMoloko (dismissed):
2 small suggestions. In general LGTM :+1:

source/user_data_v3.rst
```diff
@@ -89,3 +91,48 @@ Spinning a new node connecting to "10.0.219.209" as a seed, and installing cloud
...
+    scylla_yaml:
+      cluster_name: test-cluster
+      experimental: true
+      seed_provider:
+        - class_name: org.apache.cassandra.locator.SimpleSeedProvider
+          parameters:
+            - seeds: 10.0.219.209
+      post_configuration_script: "#! /bin/bash\nyum install cloud-init-cfn"
+      start_scylla_on_first_boot: true
+
```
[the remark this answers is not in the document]
fruch: yeah, thought of adding a more complex example

common/scylla_configure.py
fruch: seem like we don't need line 103, and the yaml can read just fine json.

fruch: removed the `json.loads`, it's not needed

source/user_data_v3.rst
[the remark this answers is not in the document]
fruch: added it

common/scylla_configure.py
fruch: fixed

README.md
vponomaryov: Not resolved

soyacz (dismissed):

KnifeyMoloko (dismissed):

common/scylla_configure.py
```diff
@@ -85,10 +84,22 @@ def save_scylla_yaml(self):
...
-                self._instance_user_data = json.loads(raw_user_data) if raw_user_data.strip() else {}
-                LOGGER.debug("JSON parsed user-data: %s", self._instance_user_data)
+
+                # Try reading mime multipart message, and extract
+                # scylla-machine-image configuration out of it
+                message = message_from_string(raw_user_data)
+                if message.is_multipart():
+                    for part in message.walk():
+                        if part.get_content_type() in ('x-scylla/json', 'x-scylla/yaml'):
+                            raw_user_data = part.get_payload()
```
vponomaryov, line 96: You have a bug here:
The `raw_user_data` gets overwritten when we have more than one part of the `json/yaml` types.

vponomaryov (changes requested):
Please, see inline comments.
Also, need to fix the commit message replacing typo for `mime`

yaronkaikov (dismissed):

common/scylla_configure.py
fruch: As a user, what would you expect, if you have passed in multiple configuration files ?

let say I've read 3 files, what should I do with them ? whom is the correct one ? do I patch them all together ?

I would rather leave it as last one wins

seems like cloud-init face a similar thing:
https://cloudinit.readthedocs.io/en/latest/topics/merging.html

I don't think we should implement anything like that now (or ever)

vponomaryov (changes requested):
We should `not merge` it while it has mentioned `BUG`.

common/scylla_configure.py
vponomaryov: If it is intentional, then, please, add an inline comment here in the code describing the intention.

vponomaryov (changes requested):
One inline comment.
Also, the topic of the commit message still has the typo for "mine".

vponomaryov (approved):
Thanks!

soyacz (approved):

yaronkaikov (approved):

PR #368, opened 2022-06-20
Drop CentOS based machine-image

syuu1228:
Since we moved to Ubuntu based machine-image, we nolonger need older
CentOS based machine-image.
Let's drop CentOS related code, and simlify build scripts.

This includes following changes:
 - Drop build_image.sh and build_with_docker.sh
 - Drop CentOS related code from scylla_install_image
 - Optimize scylla_install_image to python3 (it supports both 2 and 3 now)
 - Drop CentOS related code from scylla.json


common/scylla_image_setup
```diff
@@ -7,13 +7,10 @@
 import os
 import sys
 import pathlib
-from lib.scylla_cloud import get_cloud_instance, is_gce, is_redhat_variant
+from lib.scylla_cloud import get_cloud_instance, is_gce
```
yaronkaikov, line 10: I think we should still have the ability to install machine image on rpm based OS and run `scylla_io_setup` for example. we recently had a customer that worked on `Rhel8` and installed it manually 

Let's drop only the code related directly to building image based on rpm

yaronkaikov (changes requested):

syuu1228:
@yaronkaikov ping

yaronkaikov:
Verification passed: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/ami/1377/

yaronkaikov (approved):
LGTM

PR #373, opened 2022-07-07
scylla_install_image: add `homedir` parameter

yaronkaikov:
in https://github.com/scylladb/scylla-machine-image/pull/368/commits/69d9a7a5f855e631f6a692ed8f8270321229b6e0 `homedir` parameter was removed, causing Azure image to fail since the home dir is not `/home/ubuntu/`

Fixing it


fruch (approved):
LGTM

PR #388, opened 2022-08-02
fix yaml user-data docs

fruch:
the structure in the example was wrong, `start_scylla_on_first_boot`
and `post_configuration_script` shouldn't be under `scylla_yaml`
they should be it's siblings


benipeled (approved):
LGTM

yaronkaikov (approved):

PR #396, opened 2022-09-14
[ami]:Add parameter to select which arch to use

yaronkaikov:
Today we based our ami arch based on the local machine the process is running in. meaning that for building both x86 and arm AMI image (for same commit) we need to run on 2 different machines. Adding `arch` param so we can manually select which arch we want to build with.

default will remain the same (based on local machine)

### DIFF
--- a/packer/build_deb_image.sh
+++ b/packer/build_deb_image.sh
@@ -15,6 +15,7 @@ EXIT_STATUS=0
 DRY_RUN=false
 DEBUG=false
 TARGET=
+ARCH="$(uname -m)"
 APT_KEYS_DIR='/etc/apt/keyrings'
 APT_KEY='d0a112e067426ab2'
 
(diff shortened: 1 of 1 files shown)


yaronkaikov:
Verification:
**gce**: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/gce-image/692/
**azure**: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/azure-image/314/
**next machine image**: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/458/
**ami**: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/1600/

fruch:
@yaronkaikov what about the other clouds ? or we aren't build arm on GCP yet...

fruch (approved):
LGTM

yaronkaikov:
> @yaronkaikov what about the other clouds ? or we aren't build arm on GCP yet...

On other cloud we will use default (as today)

PR #410, opened 2022-11-20
build_image.sh:remove unrelated `if` condition

yaronkaikov:
during e19ae1250374db1541133a839ccda791593a6e6c i got some leftover `if` condition from my debuging session, not sure how did i missed it

Fixing it


benipeled (approved):

yaronkaikov:
Since the fix is still in next, I will fix it directly

PR #419, opened 2022-12-09
gce: enable multi-queue SCSI

### LINKED ISSUE #418: GCP: Enable multi-queue SCSI
See https://cloud.google.com/compute/docs/disks/optimizing-local-ssd-performance#multiqueue

1. Import the image with the VIRTIO_SCSI_MULTIQUEUE guest OS feature enabled
2. Follow the procedure to enable the module:
```
Check the value of the /sys/module/scsi_mod/parameters/use_blk_mq file.


cat /sys/module/scsi_mod/parameters/use_blk_mq
If the value of this file is Y, then multi-queue SCSI is already enabled on your imported image. If the value of the file is N, include scsi_mod.use_blk_mq=Y in the GRUB_CMDLINE_LINUX entry in your GRUB config file and restart the system.

Open the sudo nano /etc/default/grub GRUB config file in a text editor.


sudo nano /etc/default/grub
Add scsi_mod.use_blk_mq=Y to the GRUB_CMDLINE_LINUX entry.


GRUB_CMDLINE_LINUX="scsi_mod.use_blk_mq=Y"
Save the config file.

Run the update-grub command to regenerate the GRUB file and complete the configuration.


sudo update-grub
Reboot the instance.


sudo reboot
```

syuu1228:
To getting better performance on GCP instances with SCSI local disks, google cloud document recommends to enable VIRTIO_SCSI_MULTIQUEUE flag on guestOsFeatures and following kernel parameter "scsi_mod.use_blk_mq=Y".

Fixes #418

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -115,14 +115,17 @@ if __name__ == '__main__':
         setup_opt = '--ntp-domain amazon'
         sysconfig_opt = ''
         swap_opt = '--swap-directory /'
+        kernel_opt = ''
     elif args.target_cloud == 'gce':
         setup_opt = ''
         sysconfig_opt = '--disable-writeback-cache'
         swap_opt = '--swap-directory /'
+        kernel_opt = ' scsi_mod.use_blk_mq=Y'
     elif args.target_cloud == 'azure':
         setup_opt = ''
         sysconfig_opt = '--disable-writeback-cache'
         swap_opt = '--swap-directory /mnt'
+        kernel_opt = ''
 
     run('systemctl disable apt-daily-upgrade.timer apt-daily.timer dpkg-db-backup.timer motd-news.timer', shell=True, check=True)
     run('systemctl daemon-reload', shell=True, check=True)
@@ -163,7 +166,7 @@ WantedBy=multi-user.target
 
     with open('/etc/default/grub.d/50-cloudimg-settings.cfg') as f:
         grub = f.read()
-    grub = re.sub(r'^GRUB_CMDLINE_LINUX_DEFAULT="(.+)"$', r'GRUB_CMDLINE_LINUX_DEFAULT="\1 net.ifnames=0 clocksource=tsc tsc=reliable"', grub, flags=re.MULTILINE)
+    grub = re.sub(r'^GRUB_CMDLINE_LINUX_DEFAULT="(.+)"$', fr'GRUB_CMDLINE_LINUX_DEFAULT="\1 net.ifnames=0 clocksource=tsc tsc=reliable{kernel_opt}"', grub, flags=re.MULTILINE)
     with open('/etc/default/grub.d/50-cloudimg-settings.cfg', 'w') as f:
         f.write(grub)
     run('update-grub2', shell=True, check=True)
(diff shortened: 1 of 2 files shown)


yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/538

yaronkaikov (approved):
@syuu1228 Please verify that the module is actually loaded when creating an instance based on a fixed image. 
according to https://cloud.google.com/compute/docs/disks/optimizing-local-ssd-performance#multiqueue

we should be able to see an output of `Y` when running  `cat /sys/module/scsi_mod/parameters/use_blk_mq`

packer/scylla_install_image
```diff
@@ -115,14 +115,17 @@ if __name__ == '__main__':
         setup_opt = '--ntp-domain amazon'
         sysconfig_opt = ''
         swap_opt = '--swap-directory /'
+        kernel_opt = ''
     elif args.target_cloud == 'gce':
         setup_opt = ''
         sysconfig_opt = '--disable-writeback-cache'
         swap_opt = '--swap-directory /'
+        kernel_opt = ' scsi_mod.use_blk_mq=Y'
```
yaronkaikov, line 123: Based on the image created during the verification you did, i don't see this option available @syuu1228

yaronkaikov (changes requested):

mykaul:
Please check if we don't need to enable dm_mod.use_blk_mq=y if you are using device mapper multipath - and you may be (even on local disks)

mykaul:
> Please check if we don't need to enable dm_mod.use_blk_mq=y if you are using device mapper multipath - and you may be (even on local disks)

My laptop:
```
[ykaul@ykaul ~]$ sudo cat /sys/module/dm_mod/parameters/use_blk_mq
Y
```

avikivity:
> To getting better performance on GCP instances with SCSI local disks, google cloud document recommends to enable VIRTIO_SCSI_MULTIQUEUE flag on guestOsFeatures and following kernel parameter "scsi_mod.use_blk_mq=Y".
> 
> Fixes #418

Are those instances relevant? We should be using instances with nvme interfaces. Are some instance types not available with nvme?

Please check that the kernel option survives a kernel upgrade.

avikivity:
The option was removed in torvalds/linux@6a23e05c2fe3c64ec012fd81e51e3ab51e4f2f9f

syuu1228:
> Please check if we don't need to enable dm_mod.use_blk_mq=y if you are using device mapper multipath - and you may be (even on local disks)

Same as SCSI, seems like it's enabled by default.
Also, we don't use device mapper on our image.

syuu1228:
> Unlike the GCE document says, we cannot verify MQ is enabled by `cat /sys/module/scsi_mod/parameters/use_blk_mq`, since built-in module dees not apperer on /sys/module, and scsi module is built-in module on Ubuntu GCE kernel.

This description was probably wrong, since /sys/module/scsi_mod/parameters/ itself is exists and there are some other parameters available but use_blk_mq is not available.

Correct reason is, same as device-mapper as @avikivity commented at https://github.com/scylladb/scylla-machine-image/pull/419#issuecomment-1345627101, the option was removed: https://github.com/torvalds/linux/commit/f664a3cc17b7d0a2bc3b3ab96181e1029b0ec0e6
So, since linux 5.0, we always use multiqueue mode.

syuu1228:
I think we should close this since we don't need patch.

PR #421, opened 2022-12-26
lib/scylla_cloud.py: remove diskCount vs cpu restrictions

yaronkaikov:
although it's not recommended by GCP, we need to remove the restriction for minimum cpu's when using 16 local disks or more. This is so we can support one of our cloud customer which is using this configuration

Closes: https://github.com/scylladb/scylla-enterprise/issues/2543

### DIFF
--- a/lib/scylla_cloud.py
+++ b/lib/scylla_cloud.py
@@ -362,7 +362,7 @@ class gcp_instance(cloud_instance):
                 if diskCount >= 16 and self.cpu < 32:
                     logging.warning(
                         "This machine doesn't have enough CPUs for allocated number of NVMEs (at least 32 cpus for >=16 disks). Performance will suffer.")
-                    return False
+                    return True
                 if diskCount < 1:
                     logging.warning("No ephemeral disks were found.")
                     return False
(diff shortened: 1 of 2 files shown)


fruch (approved):
LGTM

yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/9/ , artifact test failed on GCP internal error (not related to this PR), re-run the test passed - https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/artifacts/job/artifacts-gce-image-test/9/


Also tested on GCP:
image
image

syuu1228:
@yaronkaikov 
It will skip rest of the test once we returns:
```
                if diskCount >= 16 and self.cpu < 32:
                    logging.warning(
                        "This machine doesn't have enough CPUs for allocated number of NVMEs (at least 32 cpus for >=16 disks). Performance will suffer.")
                    return True
```
I guess we need to run "Disk/RAM ratio" part, so it's better to drop `return` line, not retuning True.
So it should be:
```
                if diskCount >= 16 and self.cpu < 32:
                    logging.warning(
                        "This machine doesn't have enough CPUs for allocated number of NVMEs (at least 32 cpus for >=16 disks). Performance will suffer.")
                elif diskCount < 1:
                    logging.warning("No ephemeral disks were found.")
                    return False
```

yaronkaikov:
@syuu1228 Thanks,  dequeued it from next and will send a v2

PR #422, opened 2022-12-26
v2: lib/scylla_cloud.py: remove diskCount vs cpu restrictions

yaronkaikov:
although it's not recommended by GCP, we need to remove the restriction for minimum cpu's when using 16 local disks or more. This is so we can support one of our cloud customer which is using this configuration

Closes: https://github.com/scylladb/scylla-enterprise/issues/2543


syuu1228 (approved):
LGTM

yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/10/

PR #424, opened 2023-01-08
DRAFT: [images]: install auditd by default

syuu1228:
Related with scylladb/scylla-enterprise-machine-image#54


syuu1228:
Added DRAFT, since I'm not still sure we want to add auditd machine-image only or scylla repo generically.
Maybe we can drop this PR and just add auditd dependency on our scylla package (or optionally install auditd in scylla_setup).

yaronkaikov:
@syuu1228 What is the status of this one? can we close this?

syuu1228:
This implemented at https://github.com/scylladb/scylla-enterprise-machine-image/pull/72, closing.

PR #434, opened 2023-02-06
azure: show login prompt correctly

syuu1228:
Currently, Azure image does not show Scylla login prompt when logging in to ssh, since /etc/skel/.profile does not copied to ~/.profile. To fix this, we need to copy the file while build the Azure image.

### DIFF
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -192,6 +192,12 @@
       "destination": "build/",
       "direction": "download",
       "type": "file"
+    },
+    {
+        "inline": [
+            "if [ {{build_name}} = gce -o {{build_name}} = azure ]; then sudo userdel -r -f {{user `ssh_username`}}; fi"
+        ],
+        "type": "shell"
     }
   ],
   "variables": {


yaronkaikov (commented):
@syuu1228 Are we doing any copy for AWS or GCP?

benipeled:
> Currently, Azure image does not show Scylla login prompt when logging in to ssh, since /etc/skel/.profile does not copied to ~/.profile. To fix this, we need to copy the file while build the Azure image.

That's wired. It might cause by some cleanup done by azure after the creation of the `auzreuser`
I wondered about using `/etc/profile` instead of dealing with specific user(s) / provider profile files - since the 
instance is scylla dedicated machine we can have it on all users,
WDYT?

syuu1228:
Updated to the new patch which drops $SSH_USERNAME user for both Azure/GCE, tested login prompt works on both Azure/GCE.

benipeled:
@syuu1228 can't we use `cloud-init clean` for all vendors after the build?

I would expect packer to have such an option since it kind of defaults to clean images after the build (aka cloud-init / sysprep)

syuu1228:
> @syuu1228 can't we use `cloud-init clean` for all vendors after the build?

Well, it won't clean up existing user account.

> I would expect packer to have such an option since it kind of defaults to clean images after the build (aka cloud-init / sysprep)

Seems like Packer doesn't have such option.
I found there is an option to clean up ~/.ssh/authorized_keys but no option to clean up user account entirely.

benipeled (approved):

yaronkaikov (approved):

PR #444, opened 2023-04-22
[DRAFT] move cloud snitches to external python script

syuu1228:
This is scylla-machine-image part of "move cloud snitches to external python script" PR.
Currently this PR is DRAFT version since it doesn't implemented Ec2MultiRegionSnitch yet.

----

Since it's hard to debug scylla-core's cloud snitch support, we want to move them out to python script.
On the python script we will generate cassandra-rackdc.properties file, and scylla-core will read it via GossipingPropertyFileSnitch code.

See scylladb/scylladb#12306

### DIFF
--- a/common/scylla_cloud_snitch
+++ b/common/scylla_cloud_snitch
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright 2022 ScyllaDB
+#
+# SPDX-License-Identifier: Apache-2.0
+import io
+import os
+import re
+import yaml
+import configparser
+import logging
+from abc import ABCMeta, abstractmethod
+from pathlib import PurePath
+from lib.log import setup_logging
+from lib.scylla_cloud import is_ec2, is_gce, is_azure, aws_instance, gcp_instance, azure_instance
+
+LOGGER = logging.getLogger(__name__)
+
+class JavaPropertiesParser:
+    def __load(self):
+        f = io.StringIO('[global]\n{}'.format(self._data))
+        self._cfg = configparser.ConfigParser()
+        self._cfg.optionxform = str
+        self._cfg.read_file(f)
+
+    def __escape(self, val):
+        return re.sub(r'([:=])', r'\\\1', val)
+
+    def __unescape(self, val):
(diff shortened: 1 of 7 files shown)


syuu1228:
scylla-core part of the PR is at: https://github.com/scylladb/scylladb/pull/13623

syuu1228:
I implemented JavaPropertiesParser on scylla_cloud_snitch, it's modified version of sysconfig_parser, since the format is very similar.
I wanted it just for read / write cassandra-rackdc.properties, without doing regexp stuff for each read / write, since it would be hard to maintain.

But there is a pip module to parse java properties https://pypi.org/project/javaproperties/, maybe we can consider to use it.

syuu1228:
TODO: 
 - Implement Ec2MultiRegionSnitch
     - find a way to configure broadcast_address, broadcast_rpc_address from the script. I think we need to modify scylla.yaml to do that
 - Implement exponential backoff on curl(). Currently it just sleep 5 sec for each retries

common/scylla_cloud_snitch
```diff
@@ -0,0 +1,186 @@
...
+class Ec2Snitch(CloudSnitch):
+    def __init__(self):
+        super().__init__()
+        if not is_ec2():
+            raise Exception('Could not detect EC2 instance')
+        self.__aws_instance = aws_instance()
+
+    def start(self):
+        # Split "us-east-1a" or "asia-1a" into "us-east"/"1a" and "asia"/"1a".
+        splited_zone = self.__aws_instance.availability_zone().rsplit('-', 1)
```
fruch, line 140: this is only the legacy naming support

I think we should add the support for both: 
https://cassandra.apache.org/doc/latest/cassandra/configuration/cass_rackdc_file.html#ec2_naming_scheme

and consider changing the default to match cassandra 4.x

common/scylla_cloud_snitch
fruch, line 1: it's not that lot of code, but I would recommending having unittests coverage for it.

mykaul:
@avelanarius - hopefully, the drivers may be able to (re)use this functionality as well.

yaronkaikov:
@syuu1228 What is the status with this PR?

PR #446, opened 2023-04-30
(gcp):Install latest LTS kernel during image build

yaronkaikov:
Similar to 42e05cacdedcbabf1d416e32eeb2cf317b9669f5 Dauring image creation we are running `apt-get full-upgrade` which also update the kernel (added as part of
    https://github.com/scylladb/scylla-machine-image/commit/90340275b80a3a54dcfc1e5ec660481ba167d1c3),

Since we want to use LTS kernel version only, adding the kernel removal package and installation before we run `scylla_install_image`

### DIFF
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -167,7 +167,10 @@
     {
       "type": "shell",
       "inline": [
-        "if [ {{build_name}} = aws -o {{build_name}} = azure ]; then sudo apt-get update; sudo DEBIAN_FRONTEND=noninteractive apt purge -y linux-aws* linux-headers-{{build_name}}* linux-image*{{build_name}}* linux-modules*-{{build_name}}*; sudo DEBIAN_FRONTEND=noninteractive apt-get install -y linux-{{build_name}}-lts-22.04; sudo reboot ; fi"
+        "sudo apt-get update",
+        "sudo DEBIAN_FRONTEND=noninteractive apt purge -y linux-{{build_name}}* linux-headers-{{build_name}}* linux-image*{{build_name}}* linux-modules*-{{build_name}}*",
+        "sudo DEBIAN_FRONTEND=noninteractive apt-get install -y linux-{{build_name}}-lts-22.04",
+        "sudo reboot"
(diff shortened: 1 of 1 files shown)


benipeled (approved):

yaronkaikov:
@mykaul @roydahan So it seems that GCP doesn't have an LTS kernel for `Ubuntu 22.04` after all. all I could find is [this](https://packages.ubuntu.com/focal/kernel/linux-gcp-lts-20.04) which doesn't help us

Also by default, they are now using `5.19.X` kernel version, but from searching, I don't see a proper way to downgrade the kernel, so we may need to use a different approach here. I will work on it today

syuu1228:
Seems like there is linux-gcp-lts-22.04, but it is not released yet, still on "proposed" repo.
(we can see it on pkgs.org: https://ubuntu.pkgs.org/22.04/ubuntu-proposed-main-amd64/linux-gcp-lts-22.04_5.15.0.1034.30_amd64.deb.html)
Possible workaround is enabling "proposed" repo temporarily.

We can install lts kernel from proposed repo by adding following commands before `apt-get update` on [draft PR](https://github.com/scylladb/scylla-machine-image/pull/446/files):
```
# cat <<EOF >/etc/apt/sources.list.d/ubuntu-$(lsb_release -cs)-proposed.list
# Enable Ubuntu proposed archive
deb http://archive.ubuntu.com/ubuntu/ $(lsb_release -cs)-proposed restricted main multiverse univers

cat <<EOF >/etc/apt/preferences.d/proposed-updates
# Configure apt to allow selective installs of packages from proposed
Package: *
Pin: release a=$(lsb_release -cs)-proposed
Pin-Priority: 400
EOF
```
reference: https://wiki.ubuntu.com/Testing/EnableProposed


But it's testing package, other choice will be wait for official package release.

yaronkaikov:
Since don't want to use an unofficial kernel for releases we will leave it for the time being with 5.19 (for GCP only), hopefully by the time we will release 2023.1.0 it will be official

PR #450, opened 2023-05-09
ami-arm: change instance type during image build

yaronkaikov:
Started to get an error during arm ami build:
```
09:53:19  [1;31m==> aws: Error launching source instance: InsufficientInstanceCapacity: We currently do not have sufficient a1.xlarge capacity in the Availability Zone you requested (us-east-1b). Our system will be working on provisioning additional capacity. You can currently get a1.xlarge capacity by not specifying an Availability Zone in your request or choosing us-east-1c, us-east-1d.
```

Switching to more common instance type (the basic type we are also using for testing) , `im4gn.xlarge`


Annamikhlin (approved):

yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/76/ (passed the problematic stage)

PR #451, opened 2023-05-10
[DRAFT] Add python type hinting on our scripts and run type checking

syuu1228:
To reduce runtime error on our code, and also improve code stability, add python type hinting on python code, and run type checking on github CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,9 @@ jobs:
             pip install -r test-requirements.txt
             pytest ./tests
 
+      - name: mypy
+        run: mypy .
+
       - name: Build RPM (Centos:7)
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -26,6 +26,9 @@ jobs:
             pip install -r test-requirements.txt
             pytest ./tests
 
+      - name: mypy
+        run: mypy .
```
fruch, line 30: this can be an overkill, you gonna have lots and lot of things to fix to make it work.

I would recommend starting with something like ruff, slowly.

mykaul: Can we see what we actually get running it manually?

yaronkaikov:
@syuu1228 What is the status of this one?

syuu1228:
Closing, since the patch is outdated

PR #454, opened 2023-05-11
(images):create kernel-version output with cloud source

yaronkaikov:
Today when we create images we have the same file name for all cloud images.

Let's generate a cloud specific kernel filename

Ref: https://github.com/scylladb/scylla-pkg/pull/3406

### DIFF
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -195,7 +195,7 @@
       "type": "file"
     },
     {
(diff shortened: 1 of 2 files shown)


benipeled:
I'd go with one file with all versions instead of 3 different files
(seems like we already writing the info to the 00-build)
other than that lgtm

yaronkaikov:
> I'd go with one file with all versions instead of 3 different files (seems like we already writing the info to the 00-build) other than that lgtm

Since we are running in a parallel process (for next-machine-image) we need to have separate files. Although there is no need to keep it, just use it for writing it in 00-Build.txt

PR #455, opened 2023-06-04
build.yml: fix git permission bug for checkout dir

yaronkaikov:
github action is failing with the following error:
```
fatal: unsafe repository ('/scylla-machine-image' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /scylla-machine-image
fatal: unsafe repository ('/scylla-machine-image' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /scylla-machine-image
Error: Process completed with exit code 128.
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Setup python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: 3.8                         
           architecture: x64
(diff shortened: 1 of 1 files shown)


benipeled:
How subnet-base-arch will help us solve/investigate the capacity "issue"?

yaronkaikov:
> How subnet-base-arch will help us solve/investigate the capacity "issue"?

It will not, this is a solution for now so it will work. we will need to find a way to check the actual availability before, i will open an issue for that

yaronkaikov:
> > How subnet-base-arch will help us solve/investigate the capacity "issue"?
> 
> It will not, this is a solution for now so it will work. we will need to find a way to check the actual availability before, I will open an issue for that

The default subnet_id we were using until today was us-east-1b which is not good for `aarch64` (at the moment at least)

benipeled:
> > How subnet-base-arch will help us solve/investigate the capacity "issue"?
> 
> It will not, this is a solution for now so it will work. we will need to find a way to check the actual availability before, i will open an issue for that

I got confused - is there going to be a follow-up change? so why not add them together?

> The default subnet_id we were using until today was us-east-1b which is not good for aarch64 (at the moment at least)

The capacity issue at the moment, is with AZ A but it can be changed to AZ B in a few minutes/hours/days

fruch:
docs says you dont have to specify zone, or subnet...
https://developer.hashicorp.com/packer/plugins/builders/amazon/instance#availability_zone

yaronkaikov:
dropping the changes related to the subnet configuration. Let's see if waiting will actually solve the issue. 
In parallel i will look for better solution

yaronkaikov:
@fruch Please review

fruch (approved):
LGTM

fruch:
> @fruch Please review

seems like the rocky8 image got the newer git version as well

PR #465, opened 2023-07-12
packer-shell: replace semicolon separators with double ampersand

benipeled:
As part of investigating another issue (missing linux-azure-lts package) I found out that the update-kernel shell commands are set with semicolon separators which means all commands will run regardless of the results of previous commands,

This commits change it to be separated by double ampersand to make sure commands will be run only if the previous command is done successfully


benipeled:
Verified with 3 different https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/azure-image builds and 2 different https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/ builds (which for some reason didn't fail on the other issue of missing linux-azure-lts)

Annamikhlin (approved):

PR #470, opened 2023-08-15
Avoid rsyslog filling disk space for branch-5.1

syuu1228:
Default configuration of logrotate is not enough to avoid disk full when logging speed is too fast.
To make sure it rotates logs before disk full, we need to run logrotate much more frequently, and also do rotation in size based, not time based.

Related scylladb/scylla-enterprise#3080


syuu1228:
We cannot merge this to master branch, since we do not have it anymore, because we moved to Ubuntu Minimal (see more detail on https://github.com/scylladb/scylla-enterprise/issues/3080#issuecomment-1674141343).
branch-5.1 is the latest version which uses rsyslog + logrotate on the image.

yaronkaikov:
@syuu1228 Is this ready for review? please validate the change doesn't break anything

yaronkaikov (approved):
LGTM - please validate the changes before merging

PR #472, opened 2023-08-17
gce: drop rsyslog for branch-5.3

syuu1228:
On AWS and Azure Ubuntu Minimal image we don't have rsyslog, it just uses persistent journal log, but only GCE image has rsyslog. We want to align image configuration between clouds, let's drop rsyslog from GCE image.

Related scylladb/scylla-enterprise#3080

(cherry picked from commit a63350da21eddc2577604e182e73b7c54c59affb)


yaronkaikov (approved):

yaronkaikov:
Thanks @syuu1228 , Normally you can just label the original PR with the backport label and we will do it without the need of review.

PR #473, opened 2023-08-28
scylla_cloud_io_setup: add preset configuration for all supported instance types on EC2

syuu1228:
Since we merged https://github.com/scylladb/scylladb/commit/817f34d1a94a1daa1e8591c93a047081d51f5628, we are missing preset configuration for c5d, m5d, m5ad, r5d, z1d. On such instance, our AMI causes error during startup (see scylladb/scylladb#9660).
To fix the problem, we need to add preset configuration for these instance types.

### DIFF
--- a/common/scylla_cloud_io_setup
+++ b/common/scylla_cloud_io_setup
@@ -81,6 +81,172 @@ class aws_io_setup(cloud_io_setup):
             self.disk_properties["read_bandwidth"] = 507338935 * nr_disks
             self.disk_properties["write_iops"] = 57100 * nr_disks
             self.disk_properties["write_bandwidth"] = 483141731 * nr_disks
+        elif idata.instance_class() in ("m5d", "r5d"):
+            if idata.instance_size() == "large":
+                disk_properties["read_iops"] = 33271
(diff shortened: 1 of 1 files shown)


yaronkaikov (approved):

mykaul:
@syuu1228 - at some point, we should move this code into a Python dictionary, no?

common/scylla_cloud_io_setup
```diff
@@ -81,6 +81,172 @@ class aws_io_setup(cloud_io_setup):
             self.disk_properties["read_bandwidth"] = 507338935 * nr_disks
             self.disk_properties["write_iops"] = 57100 * nr_disks
             self.disk_properties["write_bandwidth"] = 483141731 * nr_disks
+        elif idata.instance_class() in ("m5d", "r5d"):
```
benipeled, line 84: @syuu1228 it should be `self.idata.instance_class()` - same for following refs ofc

benipeled:
@syuu1228 please make sure to verify every change in advanced
In this case, use the machine-image releng-testing pipeline (or byo)

PR #476, opened 2023-08-30
Add missing i4g.4xlarge instnace type

yaronkaikov:
Adding pre-configure io_setup values
Also since `i4g` doesn't support enhanced networking adding it to use ENA

### DIFF
--- a/common/scylla_cloud_io_setup
+++ b/common/scylla_cloud_io_setup
@@ -308,6 +308,11 @@ class aws_io_setup(cloud_io_setup):
             self.disk_properties["read_bandwidth"] = 1148509696
             self.disk_properties["write_iops"] = 82746
             self.disk_properties["write_bandwidth"] = 508828810
+        elif self.idata.instancetype == "i4g.4xlarge":
+            self.disk_properties["read_iops"] = 272704
+            self.disk_properties["read_bandwidth"] = 2297019392
+            self.disk_properties["write_iops"] = 165492
+            self.disk_properties["write_bandwidth"] = 1017657620
         elif self.idata.instancetype == "i4g.8xlarge":
             self.disk_properties["read_iops"] = 271495 * nr_disks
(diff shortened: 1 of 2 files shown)


Annamikhlin:
Please check if you need to remove the `i4g.large` type, if it not should be supported 
https://github.com/scylladb/scylla-machine-image/blob/a5c370764b976059a089e888b849115504a1c8ee/common/scylla_cloud_io_setup#L296-L300

yaronkaikov:
> Please check if you need to remove the `i4g.large` type, if it not should be supported
> 
> https://github.com/scylladb/scylla-machine-image/blob/a5c370764b976059a089e888b849115504a1c8ee/common/scylla_cloud_io_setup#L296-L300

We should support it as well. so keeping it

Annamikhlin:
> > Please check if you need to remove the `i4g.large` type, if it not should be supported
> > https://github.com/scylladb/scylla-machine-image/blob/a5c370764b976059a089e888b849115504a1c8ee/common/scylla_cloud_io_setup#L296-L300
> 
> We should support it as well. so keeping it

So we are supporting it but not testing it?

Annamikhlin:
> > > Please check if you need to remove the `i4g.large` type, if it not should be supported
> > > https://github.com/scylladb/scylla-machine-image/blob/a5c370764b976059a089e888b849115504a1c8ee/common/scylla_cloud_io_setup#L296-L300
> > 
> > 
> > We should support it as well. so keeping it
> 
> So we are supporting it but not testing it?

Ok, I see your answer in https://github.com/scylladb/scylla-pkg/pull/3556#discussion_r1309918566 :)

Annamikhlin (approved):
lgtm

benipeled (approved):

yaronkaikov:
Verified with https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/artifacts/job/artifacts-ami-arm-test/71/

PR #478, opened 2023-09-13
[azure]:fix resource cleanup retry

yaronkaikov:
During azure image build we keep seeing:
```
07:39:44  ␛[1;32m==> azure: Error deleting resource. Will retry.
07:39:44  ==> azure: Name: pkrip4th7si0bxq
07:39:44  ==> azure: Error: network.PublicIPAddressesClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="PublicIPAddressCannotBeDeleted" Message="Public IP address /subscriptions/****/resourceGroups/scylla-images/providers/Microsoft.Network/publicIPAddresses/pkrip4th7si0bxq can not be deleted since it is still allocated to resource /subscriptions/****/resourceGroups/scylla-images/providers/Microsoft.Network/networkInterfaces/pkrni4th7si0bxq/ipConfigurations/ipconfig. In order to delete the public IP, disassociate/detach the Public IP address from the resource.  To learn how to do this, see aka.ms/deletepublicip." Details=[]
```

Closes: https://github.com/scylladb/scylla-pkg/issues/3021

### DIFF
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -168,6 +168,7 @@
       },
       "vm_size": "{{user `vm_size`}}",
       "build_resource_group_name": "scylla-images",
+      "async_resourcegroup_delete": true,
       "keep_os_disk": true,
       "virtual_network_name": "scylla-images",
       "private_virtual_network_with_public_ip": true


packer/scylla.json
```diff
@@ -168,6 +168,7 @@
       },
       "vm_size": "{{user `vm_size`}}",
       "build_resource_group_name": "scylla-images",
+      "async_resourcegroup_delete": false,
```
benipeled, line 171: You probably meant to set it `true`
To be on the safe side - please add a calendar/slack alert  for two months from now to check that we have no left overs resources on Azure caused by this change

yaronkaikov:
I ran https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/azure-image/194/consoleFull and verified it's working, and I don't see this error. But all other runs show the same error. So i am not sure this change is doing anything.

yaronkaikov:
Closing this. as there is no solution for those retry (at least nothing  i could find)

PR #483, opened 2023-09-23
Optimize cloud detection

### LINKED ISSUE #481: Long delay after SSH login
When logging into a VM created from our machine-image, there is a significant delay of approximately 30 seconds, after the first part of the scylla_login script, I haven't checked it yet but I think that most time consumed by gathering the cloud-instance obj
https://github.com/scylladb/scylla-machine-image/blob/16b454e76f2016b33e7a0fd3959166c61d4a3283/common/scylla_login#L113

I saw it on GCP and AWS but it's probably the same case with Azure

image

syuu1228:
Currently get_cloud_instance() is very slow on GCE / Azure, since it fails to detect AWS metadata server, and retrying multiple times in curl().

Actually we don't really need to do detection, we can specify target cloud name statically when we build machine-image.

To do so, adding target cloud name to machine_image_info.yaml while building machine-image, and reference it from scylla_cloud.py.

Note that we still need to fallback to metadata server based detection, since there are users who manually install scylla-machine-image package , not using our machine-image.

Fixes #481

### DIFF
--- a/lib/scylla_cloud.py
+++ b/lib/scylla_cloud.py
@@ -18,6 +18,7 @@ import glob
 import distro
 import base64
 import datetime
+import asyncio
 from subprocess import run, CalledProcessError
 from abc import ABCMeta, abstractmethod
 
@@ -58,23 +59,43 @@ def scylla_excepthook(etype, value, tb):
 sys.excepthook = scylla_excepthook
 
 
+def _curl_one(url, headers=None, method=None, byte=False, timeout=3):
+    req = urllib.request.Request(url, headers=headers or {}, method=method)
+    with urllib.request.urlopen(req, timeout=timeout) as res:
+        if byte:
+            return res.read()
+        else:
+            return res.read().decode('utf-8')
+
 # @param headers dict of k:v
 def curl(url, headers=None, method=None, byte=False, timeout=3, max_retries=5, retry_interval=5):
     retries = 0
     while True:
         try:
-            req = urllib.request.Request(url, headers=headers or {}, method=method)
-            with urllib.request.urlopen(req, timeout=timeout) as res:
-                if byte:
-                    return res.read()
-                else:
-                    return res.read().decode('utf-8')
+            return _curl_one(url, headers, method, byte, timeout)
         except (urllib.error.URLError, socket.timeout):
             time.sleep(retry_interval)
             retries += 1
             if retries >= max_retries:
                 raise
 
+async def aiocurl(url, headers=None, method=None, byte=False, timeout=3, max_retries=5, retry_interval=5):
+    retries = 0
+    while True:
+        try:
+            return _curl_one(url, headers, method, byte, timeout)
+        except (urllib.error.URLError, socket.timeout):
+            await asyncio.sleep(retry_interval)
+            retries += 1
+            if retries >= max_retries:
+                raise
+
+def read_one_line(filename):
+    try:
+        with open(filename) as f:
+            return f.read().strip()
+    except FileNotFoundError:
+        return ''
 
 class cloud_instance(metaclass=ABCMeta):
     @abstractmethod
@@ -126,6 +147,20 @@ class cloud_instance(metaclass=ABCMeta):
     def endpoint_snitch(self):
         pass
 
+    @classmethod
+    @abstractmethod
+    def identify_dmi(cls):
+        pass
+
+    @classmethod
+    @abstractmethod
+    async def identify_metadata(cls):
+        pass
+
+    @classmethod
+    async def identify(cls):
+        return cls.identify_dmi() or await cls.identify_metadata()
+
 
 
 class gcp_instance(cloud_instance):
@@ -150,13 +185,20 @@ class gcp_instance(cloud_instance):
         return self.ENDPOINT_SNITCH
 
 
-    @staticmethod
-    def is_gce_instance():
+    @classmethod
+    def identify_dmi(cls):
+        product_name = read_one_line('/sys/class/dmi/id/product_name')
+        if product_name == "Google Compute Engine":
+            return cls
+        return None
+
+    @classmethod
+    async def identify_metadata(cls):
         """Check if it's GCE instance via DNS lookup to metadata server."""
         try:
             addrlist = socket.getaddrinfo('metadata.google.internal', 80)
         except socket.gaierror:
-            return False
+            return None
         for res in addrlist:
             af, socktype, proto, canonname, sa = res
             if af == socket.AF_INET:
@@ -164,11 +206,13 @@ class gcp_instance(cloud_instance):
                 if addr == "169.254.169.254":
                     # Make sure it is not on GKE
                     try:
-                        gcp_instance().__instance_metadata("machine-type")
+                        await aiocurl(cls.META_DATA_BASE_URL + "machine-type?recursive=false",
+                    headers={"Metadata-Flavor": "Google"})
                     except urllib.error.HTTPError:
-                        return False
-                    return True
-        return False
+                        return None
+                    return cls
+        return None
+
 
     def __instance_metadata(self, path, recursive=False):
         return curl(self.META_DATA_BASE_URL + path + "?recursive=%s" % str(recursive).lower(),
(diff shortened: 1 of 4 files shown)


syuu1228:
Tested at https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/next-machine-image/196/

mykaul:
nit: you should be able to reliably detect cloud based on DMI data - from https://coroot.com/blog/cloud-metadata : 
```
AWS Xen instances: the uuid from /sys/hypervisor/uuid has the "ec2" or "EC2" prefix
AWS Nitro instances: the content of /sys/class/dmi/id/board_vendor is "Amazon EC2"
GCP: the content of /sys/class/dmi/id/board_vendor is "Google"
Azure: the content of /sys/class/dmi/id/board_vendor is "Microsoft Corporation"
```

syuu1228:
> nit: you should be able to reliably detect cloud based on DMI data - from https://coroot.com/blog/cloud-metadata :
> 
> ```
> AWS Xen instances: the uuid from /sys/hypervisor/uuid has the "ec2" or "EC2" prefix
> AWS Nitro instances: the content of /sys/class/dmi/id/board_vendor is "Amazon EC2"
> GCP: the content of /sys/class/dmi/id/board_vendor is "Google"
> Azure: the content of /sys/class/dmi/id/board_vendor is "Microsoft Corporation"
> ```

I also considered that, it would works well on AWS / GCP, but on Azure DMI conteints are exactly same as baremetal Hyper-V.
To cause correct error message "Unknown cloud provider! Only AWS/GCP/Azure supported" on baremetal Hyper-V, we still need to keep metadata server access on `azure_instance.is_azure_instance()`, so we cannot drop network access entirely from cloud detection.

see: https://stackoverflow.com/questions/11570965/how-to-detect-azure-amazon-vm

Maybe we can wait metadata server access just for Azure, since it will likely success because in that case we already know it's not AWS nor GCP.

Or, maybe we can ignore about non-cloud environment, since scylla-machine-image script will does not work on such environment at some point.
But it will not clealer message I mentioned above, it will cause Traceback at some point instead.

lib/scylla_cloud.py
```diff
@@ -823,25 +825,65 @@ def user_data(self):
...
-    return aws_instance.is_aws_instance()
+    return cloud_detector.get_instance().is_ec2()
 
 def is_gce():
-    return gcp_instance.is_gce_instance()
+    return cloud_detector.get_instance().is_gce()
 
 def is_azure():
-    return azure_instance.is_azure_instance()
+    return cloud_detector.get_instance().is_azure()
```
benipeled, line 883: I'm not sure these is_ec2/is_gce/is_azure functions are required with this new implementation adding the `cloud_detector()` class,
We can call the cloud_detector.is_ec2 directly on the required places

mykaul:
> > nit: you should be able to reliably detect cloud based on DMI data - from https://coroot.com/blog/cloud-metadata :
> > ```
> > AWS Xen instances: the uuid from /sys/hypervisor/uuid has the "ec2" or "EC2" prefix
> > AWS Nitro instances: the content of /sys/class/dmi/id/board_vendor is "Amazon EC2"
> > GCP: the content of /sys/class/dmi/id/board_vendor is "Google"
> > Azure: the content of /sys/class/dmi/id/board_vendor is "Microsoft Corporation"
> > ```
> 
> I also considered that, it would works well on AWS / GCP, but on Azure DMI conteints are exactly same as baremetal Hyper-V. To cause correct error message "Unknown cloud provider! Only AWS/GCP/Azure supported" on baremetal Hyper-V, we still need to keep metadata server access on `azure_instance.is_azure_instance()`, so we cannot drop network access entirely from cloud detection.
> 
> see: https://stackoverflow.com/questions/11570965/how-to-detect-azure-amazon-vm
> 
> Maybe we can wait metadata server access just for Azure, since it will likely success because in that case we already know it's not AWS nor GCP.
> 
> Or, maybe we can ignore about non-cloud environment, since scylla-machine-image script will does not work on such environment at some point. But it will not clealer message I mentioned above, it will cause Traceback at some point instead.

We can ignore Hyper-V, or treat it just like Azure - both cases are fine, IMHO.

fruch:
We could do something similar to

https://github.com/dgzlopes/cloud-detect/blob/master/cloud_detect/providers/azure_provider.py

mykaul:
> We could do something similar to
> 
> https://github.com/dgzlopes/cloud-detect/blob/master/cloud_detect/providers/azure_provider.py

This needs the networking stack to be up (and IPv4 configured, what about IPv6 only?) - we'd like to avoid that.

avikivity:
There are also metal instances, which might be different.

mykaul:
> There are also metal instances, which might be different.

That could be the fallback if above detection fails.

syuu1228:
Pushed new implementation which uses DMI for the detection, and async/await for metadata server access (when DMI probe failed).
The implementation referenced https://github.com/dgzlopes/cloud-detect, borrowed some ideas:
 - which DMI parameter to detect the cloud
 - order of access DMI and metadata server (try DMI first, then metadata server)
 - use async/await for metadata server access

syuu1228:
Ah, it breaking the unittest, fixing...

syuu1228:
Testcase fixed.

fruch (approved):
LGTM

even that I would prefer using https://github.com/dgzlopes/cloud-detect directly
and not copying it's implementation (even it's MIT license, and we could do it legally)

syuu1228:
> LGTM
> 
> even that I would prefer using https://github.com/dgzlopes/cloud-detect directly and not copying it's implementation (even it's MIT license, and we cloud do it legally)

We actually can package cloud-detect package into scylla-python3 by updating install-dependencies.sh in scylla repo, but I did not prefer to update that just for scylla-machine-image so I haven't added any package for scylla-machine-image scripts.
That's the reason why I haven't used cloud-detect in this PR, and also other reason is we already have metadata based detection and DMI based one should be very small change.

But if we copy scylla-cqlsh implementation which adds zipped pip package into relocatable package, then we can use pip packages on scylla-machine-image too, that would be alternative way to implement this.

fruch:
> > LGTM
> > 
> > even that I would prefer using https://github.com/dgzlopes/cloud-detect directly and not copying it's implementation (even it's MIT license, and we cloud do it legally)
> 
> We actually can package cloud-detect package into scylla-python3 by updating install-dependencies.sh in scylla repo, but I did not prefer to update that just for scylla-machine-image so I haven't added any package for scylla-machine-image scripts.
> That's the reason why I haven't used cloud-detect in this PR, and also other reason is we already have metadata based detection and DMI based one should be very small change.
> 
> But if we copy scylla-cqlsh implementation which adds zipped pip package into relocatable package, then we can use pip packages on scylla-machine-image too, that would be alternative way to implement this.

A simple vendoring of the code as a submodule, might also work in this case.

syuu1228:
> > > even that I would prefer using https://github.com/dgzlopes/cloud-detect directly and not copying it's implementation (even it's MIT license, and we cloud do it legally)
> > 
> > 
> > We actually can package cloud-detect package into scylla-python3 by updating install-dependencies.sh in scylla repo, but I did not prefer to update that just for scylla-machine-image so I haven't added any package for scylla-machine-image scripts.
> > That's the reason why I haven't used cloud-detect in this PR, and also other reason is we already have metadata based detection and DMI based one should be very small change.
> > But if we copy scylla-cqlsh implementation which adds zipped pip package into relocatable package, then we can use pip packages on scylla-machine-image too, that would be alternative way to implement this.
> 
> A simple vendoring of the code as a submodule, might also work in this case.

Tried to add cloud-detect as submodule and load from scylla-machine-image, but it seems difficult because the module is hevily depend with aiohttp, witch contains C extension.
Probably we should just update scylla-python3 dependencies to add the module since it's simplest solution.

avikivity:
Is there no simple way to do this without adding complicated packages?

syuu1228:
> Is there no simple way to do this without adding complicated packages?

I think current patch on this PR does this.
It referenced cloud-detect, borrowed some ideas, but not adding additional packages:
https://github.com/scylladb/scylla-machine-image/pull/483#issuecomment-1742664119

syuu1228:
@yaronkaikov Tested on https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/264/, please merge this.

yaronkaikov (approved):

PR #487, opened 2023-10-18
aws: use yaml module for cloud.cfg

syuu1228:
On latest master, re.sub() is failing since regex pattern is not matched with cloud.cfg on latest Ubuntu image.
To avoid such problem, we should use yaml module instead.

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -14,6 +14,7 @@ import shlex
 import tarfile
 import argparse
 import platform
+import yaml
 from io import StringIO
 from subprocess import run, PIPE, STDOUT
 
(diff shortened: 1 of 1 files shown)


syuu1228:
Note that, once we rewrite YAML file with yaml.dump(), we lose all comments on the file.
That was the reason why we currently using re.sub() instead of yaml module.
But I think it’s more important to avoid such build failure.

fruch:
> Note that, once we rewrite YAML file with yaml.dump(), we lose all comments on the file. That was the reason why we currently using re.sub() instead of yaml module. But I think it’s more important to avoid such build failure.

we have the same problem with scylla.yaml as well.

we probably should swith to https://pypi.org/project/ruamel.yaml/ which does preserve comments

yaronkaikov:
@syuu1228 Running https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/234/ to verify it's working. once it pass i will merge

yaronkaikov (approved):

yaronkaikov:
> @syuu1228 Running https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/234/ to verify it's working. once it pass i will merge

Passed the problematic phase

syuu1228:
> we have the same problem with scylla.yaml as well.
> 
> we probably should swith to https://pypi.org/project/ruamel.yaml/ which does preserve comments

@fruch Yes, maybe we should try that.

PR #488, opened 2023-10-19
adding EC2 instance type option to use while building the AMI + change current default 

Annamikhlin:
This PR includes 2 commits:
- 780dd8753c90fad1b6933366fb64caf9bf8260c7 - change default INSTANCE_TYPE for ARM to im4gn.2xlarge
- 1bdf38e781ab40780d271a6825c90395fa51a723 - adding new 'ec2-instance-type' parameter

### DIFF
--- a/packer/build_image.sh
+++ b/packer/build_image.sh
@@ -274,7 +274,7 @@ if [ "$TARGET" = "aws" ]; then
         ;;
       "aarch64")
         SOURCE_AMI_FILTER="ubuntu-minimal/images/hvm-ssd/ubuntu-jammy-22.04-arm64*"
-        INSTANCE_TYPE="im4gn.xlarge"
+        INSTANCE_TYPE="im4gn.2xlarge"
         ;;
       *)
         echo "Unsupported architecture: $arch"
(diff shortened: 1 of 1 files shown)


Annamikhlin:
Verification: 
https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/238/ - running with default `instance_type=im4gn.2xlarge`

packer/build_image.sh
```diff
@@ -152,6 +153,10 @@ while [ $# -gt 0 ]; do
             ARCH="$2"
             shift 2
             ;;
+        "--ec2-instance-type")
```
yaronkaikov, line 156: Maybe we could do general `instance-type` since we are not building all targets at once anyway. I would rename it to `instance-type`  and use it for all targets 
Other than that LGTM

benipeled (approved):

packer/build_image.sh
Annamikhlin: I put it it with EC2 naming because today it is relevant only for AWS, we are using this parameter only in case
https://github.com/scylladb/scylla-machine-image/blob/1bdf38e781ab40780d271a6825c90395fa51a723/packer/build_image.sh#L268

Issue #495, opened 2023-12-14
AWS: use gp3 volume_type for more performance and lower cost


yaronkaikov:
As default, all our images (and snapshots) are `gp2` type. 
We should move all to `gp3` which is more cost effective and better performance

Refs: https://cloudfix.com/blog/migrate-gp2-to-gp3-better-performance-lower-costs/

/cc @mykaul

PR #496, opened 2023-12-14
ami: create with gp3 as volume_type

yaronkaikov:
Until today,  all our images (and snapshots) are gp2 type.

We should move all to gp3 which is more cost effective and better performance

Closes: https://github.com/scylladb/scylla-machine-image/issues/495


yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/ami/286/
image

mykaul (approved):

PR #500, opened 2023-12-24
gpg: add the 2024 pub key

benipeled:
...and remove the old apt-get which is no longer required

Ref https://github.com/scylladb/scylla-pkg/issues/3738

### DIFF
--- a/packer/build_image.sh
+++ b/packer/build_image.sh
@@ -15,7 +15,7 @@ DEBUG=false
 BUILD_MODE='release'
 TARGET=
 APT_KEYS_DIR='/etc/apt/keyrings'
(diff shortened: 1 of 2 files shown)


fruch:
@benipeled we need this also in SCT, right ?

benipeled:
> @benipeled we need this also in SCT, right ?

Correct, SCT is on the list to be updated among other places

benipeled:
Verified by https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/265/
Include artifacts-{aws,gce,azure}-tests

yaronkaikov (approved):

PR #504, opened 2024-01-31
images: clean old and un-used code

yaronkaikov:
since they not in use anymore removing them


yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/275/

benipeled (approved):

PR #507, opened 2024-02-06
aws_instance: fix broken user_data()

### LINKED ISSUE #498: Cannot get EC2 user data after enabled IMDSv2
`Dec 19 15:38:09 ip-172-19-20-73 scylla_post_start.py[3760]: 2023-12-19 15:38:09,968 - [user_data] - WARNING - Error getting user data: HTTP Error 401: Unauthorized. Will use defaults!`

`Dec 19 15:38:09 ip-172-19-20-73 scylla_post_start.py[3760]: Error getting user data: HTTP Error 401: Unauthorized. Will use defaults!`

the scylla_post_start.py script failed to fetch user data after enabled IMDSv2. It needs to fetch a IMDSv2 token https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

syuu1228:
user_data() is mistakenly unsupported IMDSv2, since it's not using __instance_metadata().

Fixed the code to use __instance_metadata().

Fixes #498

### DIFF
--- a/common/scylla_configure.py
+++ b/common/scylla_configure.py
@@ -115,7 +115,10 @@ class ScyllaMachineImageConfigurator(UserData):
             try:
                 default_timeout = self.CONF_DEFAULTS["post_configuration_script_timeout"]
                 script_timeout = self.instance_user_data.get("post_configuration_script_timeout", default_timeout)
-                decoded_script = base64.b64decode(post_configuration_script)
+                try:
+                    decoded_script = base64.b64decode(post_configuration_script)
+                except binascii.Error:
+                    decoded_script = post_configuration_script
                 LOGGER.info("Running post configuration script:\n%s", decoded_script)
                 subprocess.run(decoded_script, check=True, shell=True, timeout=int(script_timeout))
             except Exception as e:
--- a/lib/scylla_cloud.py
+++ b/lib/scylla_cloud.py
@@ -696,7 +696,7 @@ class aws_instance(cloud_instance):
             time_diff_sec = int(time_diff.total_seconds())
             if time_diff_sec >= self.METADATA_TOKEN_TTL - 120:
                 self.__refresh_metadata_token()
-        return curl(self.META_DATA_BASE_URL + "meta-data/" + path, headers={"X-aws-ec2-metadata-token": self._metadata_token})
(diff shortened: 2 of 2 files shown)


syuu1228:
Tested at https://jenkins.scylladb.com/view/releng-testing/job/scylla-master/job/releng-testing/job/next-machine-image/280/

yaronkaikov (approved):
LGTM

syuu1228:
@yaronkaikov The patch is already merged though, I have one thing don't understand.

When I testing this patch with IMDSv2 mode userdata, I used example code of the userdata from our README.md.
Witch is something like this:
```
scylla_yaml:
  cluster_name: test-cluster
  experimental: true
  seed_provider:
    - class_name: org.apache.cassandra.locator.SimpleSeedProvider
      parameters:
        - seeds: 10.0.219.209
post_configuration_script: "#! /bin/bash\nyum install cloud-init-cfn"
start_scylla_on_first_boot: true
```
(The script was for CentOS7 so I modified the command to apt-get, but anyway)

But I realized that it causes "binascii.Error: Incorrect padding" exception while loading post_configuration_script.
Because our script expected post_configuration_script is base64 encoded, but the example YAML code doesn't.

So it only accepts something like this:
```
scylla_yaml:
  cluster_name: test-cluster
  experimental: true
  seed_provider:
    - class_name: org.apache.cassandra.locator.SimpleSeedProvider
      parameters:
        - seeds: 10.0.219.209
post_configuration_script: IyEgL2Jpbi9iYXNoCnl1bSBpbnN0YWxsIGNsb3VkLWluaXQtY2Zu # base64 encoded
start_scylla_on_first_boot: true
```

Entire userdata will encode in base64 anyway (at least on EC2 I think), so post_configuration_script will encoded twice.
I'm not sure this is correct behavior since our example code does not have such thing, so I added try-except condition on my patch to both:
https://github.com/scylladb/scylla-machine-image/pull/507/files#diff-71a59e03175d0b8ae7712c9a6ddc7bc6daeb902291a81b2f40ef59dda4d344f9R118

But I actually found both SCT and testcase in scylla-machine-image expected post_configuration_script is encoded in base64.

Do you know which way is correct?
If base64 encoded is the correct way, I will send new patch to drop my changes.

PR #518, opened 2024-04-04
packer: share AMIs with the entire aws scylla organization

benipeled:
With this change, AMIs wil be shared with the entire aws scylla org (instead of specific accounts) to allow all accounts to use the AMI before it is promoted + avoid the need to maintain a list of accounts that require AMI access,

This is about the AMI, on another change on pkg we need to adjust the `change_snapshot_permissions()` function to loop through all the org account

Ref https://github.com/scylladb/scylla-pkg/issues/3529

### DIFF
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -95,7 +95,10 @@
         "delay_seconds": "30",
         "max_attempts": "100"
       },
-      "shutdown_behavior": "terminate"
+      "shutdown_behavior": "terminate",
+      "ami_org_arns": [
+          "arn:aws:organizations::978072043225:organization/o-o561yy1rs6"
+      ]
     },
     {
       "name": "gce",


benipeled:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/293/

yaronkaikov (approved):

benipeled:
> AFAIK the cloud needs those images on the enterprise branch for testing, I don't see a reason to backport this.

Yes, and they probably need it for future releases only but since it's a minor trivial change we'll backport it

yaronkaikov:
> > AFAIK the cloud needs those images on the enterprise branch for testing, I don't see a reason to backport this.
> 
> Yes, and they probably need it for future releases only but since it's a minor trivial change we'll backport it

We should backport if it's needed or if it's improved something (even if it's a minor and there is no risk).  Let's verify before if it's needed for releases. Thanks

benipeled:
> > > AFAIK the cloud needs those images on the enterprise branch for testing, I don't see a reason to backport this.
> > 
> > 
> > Yes, and they probably need it for future releases only but since it's a minor trivial change we'll backport it
> 
> We should backport if it's needed or if it's improved something (even if it's a minor and there is no risk). Let's verify before if it's needed for releases. Thanks

Yes it's an improvement and I'm not sure what verification you're looking for here,
In any case, let's please avoid a long discussion about something so minor that at the end will effect/help nothing

yaronkaikov:
> > > > AFAIK the cloud needs those images on the enterprise branch for testing, I don't see a reason to backport this.
> > > 
> > > 
> > > Yes, and they probably need it for future releases only but since it's a minor trivial change we'll backport it
> > 
> > 
> > We should backport if it's needed or if it's improved something (even if it's a minor and there is no risk). Let's verify before if it's needed for releases. Thanks
> 
> Yes it's an improvement and I'm not sure what verification you're looking for here, In any case, let's please avoid a long discussion about something so minor that at the end will effect/help nothing

My point exactly to will help nothing :-)

Annamikhlin (approved):

PR #519, opened 2024-04-22
Mount separated /tmp, /var/tmp partition with noexec, nodev, nosuid options

syuu1228:
Mount separated /tmp, /var/tmp partition with apply noexec, nodev, nosuid options.

For /tmp it will use tmpfs.
For /var/tmp, since it's difficult to add separated partition on machine-image,
it will add 1GB loopback image on /vartmpfile, just like /swapfile.
To setup the file, added scylla_var_tmp_setup which is modified version
of scylla_swap_setup.

- xccdf_org.ssgproject.content_rule_partition_for_tmp
- xccdf_org.ssgproject.content_rule_mount_option_tmp_nodev
- xccdf_org.ssgproject.content_rule_mount_option_tmp_noexec
- xccdf_org.ssgproject.content_rule_mount_option_tmp_nosuid
- xccdf_org.ssgproject.content_rule_mount_option_var_tmp_nodev
- xccdf_org.ssgproject.content_rule_mount_option_var_tmp_noexec
- xccdf_org.ssgproject.content_rule_mount_option_var_tmp_nosuid

Fixes scylladb/scylla-enterprise-machine-image#69
Related https://github.com/scylladb/scylla-pkg/issues/2953

### DIFF
--- a/common/scylla_var_tmp_setup
+++ b/common/scylla_var_tmp_setup
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019-present ScyllaDB
+#
+
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import os
+import sys
+import argparse
+import psutil
+from pathlib import Path
+from subprocess import run
+
+def GB(n):
+    return n * 1024 * 1024 * 1024
+
+def to_GB(n):
+    return '{:.2f}'.format(n / 1024 / 1024 / 1024)
+
+def find_mount_point(path):
+    path = path.absolute()
+    while not path.is_mount():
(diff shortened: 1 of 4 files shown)


syuu1228:
Note that why this PR does not use tmpfs for /var/tmp is, since applications may expected temporary files on /var/tmp should be preserved between reboots (reference: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard).

syuu1228:
And this is part of: https://github.com/scylladb/scylla-pkg/issues/2953
Since CIS compliance rules says /tmp and /var/tmp mount option should be noexec, nodev, nosuid.

syuu1228:
BTW, the issue is opened at enterprise version, should we move this to enterprise? Or it's fine to merge OSS version?

PR #533, opened 2024-07-29
packer/build_image.sh: update GPG key

yaronkaikov:
A new GPG key was created for supporting both rpm and deb in RSA mode

Related to https://github.com/scylladb/scylla-enterprise/issues/4222


fruch (approved):
LGTM

PR #534, opened 2024-08-05
packer/scylla_install_image: update gpg key inisde the image

yaronkaikov:
Following the changes in 23bc340, it seems i missed a place for the gpg key replacement

adding it so we can run `apt-get update` inside our images

Fixes: https://github.com/scylladb/scylla-pkg/issues/4255


yaronkaikov:
> LGTM
> 
> Is there a way to prevent this in the future? Some kind of unit test, or just one place to update these keys?

We should, and we plan

For now we are following the procedure in https://github.com/scylladb/scylla-pkg/wiki/GPG---keys-lifecycle-and-setup , i have made it more clear for next time (it was my first time :-) )

PR #536, opened 2024-09-10
scylla.yaml.j2: replace instance type from `i3` to `i4i`

yaronkaikov:
Switching to newer instance types with better performance

Fixes: https://github.com/scylladb/scylla-pkg/issues/4322


yaronkaikov:
> @yaronkaikov - I would choose the i4i.2xl as default, and you are missing the i4i.32xl

Fixed

PR #541, opened 2024-09-17
[machine-image] add backport automation

yaronkaikov:
Similar to what we have in Scylla-pkg, adding relevant automation scripts and workflow so we can easly do backports


Annamikhlin (approved):
LGTM

PR #542, opened 2024-09-18
[.github/scripts/auto-backport.py] improve label logic

yaronkaikov:
In this commit there few improvments to the backport automation process based on the work done in https://github.com/scylladb/scylla-machine-image/pull/541

1. Create new parameter `--commit` to be use only when `--pull-request` is provided
2. Trigger workflow when label is `backport/x.y` only, rather then `backport/` as it was defined before. This way we will not trigger to flow when `backport/x.y-done` was added
3. when running on specific PR with `--pull-request` make sure we don't pass an empty commit
4. fix `get_prs_from_commits()` while trying to do `get_commit` on the wrong sha1 value

### DIFF
--- a/.github/scripts/auto-backport.py
+++ b/.github/scripts/auto-backport.py
@@ -3,6 +3,7 @@
 import argparse
 import os
 import re
+import sys
 import tempfile
 import logging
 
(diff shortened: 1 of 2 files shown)


Annamikhlin (approved):

fruch:
@yaronkaikov are you gonna apply those on dtest/sct as well ?

yaronkaikov:
> @yaronkaikov are you gonna apply those on dtest/sct as well ?

yep, soon

PR #543, opened 2024-09-19
[github/workflows/add-label-when-promoted] add support for running in enterprise repo

yaronkaikov:
Adding an enterprise branch as a valid option to be triggered. This will be used for supporting backports for enterprise features


Annamikhlin (approved):
LGTM

Issue #544, opened 2024-09-23
Refactor aws_io_setup class - specifically, the IO values per instance


mykaul:
My eyes hurt from seeing so many `elif'. See https://github.com/scylladb/scylla-machine-image/blob/ad01b2a28092b0488f5de443da51a51285c98a14/common/scylla_cloud_io_setup#L41 
We should store this in a dictionary, JSON or whatnot. Single location preferably.

mykaul:
Part of this (I should open a separate issue for it!) is that we need to validate the numbers are still real. Ref https://github.com/scylladb/scylla-enterprise/issues/4713#issuecomment-2367619463
We do have a job that is testing a wide range of instances, it should probably compare the results and the hard-coded values (CC @roydahan )

roydahan:
I created a separate issue for the above comment in qa-tasks.

PR #561, opened 2024-12-05
aws: add i7ie instance type

### LINKED ISSUE #559: AWS i7ie enablement
Add pre-measured io_properties.yaml for i7ie.

https://aws.amazon.com/blogs/aws/now-available-storage-optimized-amazon-ec2-i7ie-instances/

syuu1228:
Adding preset io parameters of i7ie to scylla_cloud_io_setup, and also added i7ie to supported instance type on aws_instance class.

Closes #559

### DIFF
--- a/common/aws_io_params.yaml
+++ b/common/aws_io_params.yaml
@@ -558,3 +558,23 @@ i4i.metal:
   read_bandwidth: 3088599296
   write_iops: 239549
   write_bandwidth: 2302438912
+i7ie.large:
+  read_iops: 58449
+  read_bandwidth: 574854656
+  write_iops: 47145
+  write_bandwidth: 253132917
+i7ie.xlarge:
+  read_iops: 117257
+  read_bandwidth: 1148572714
+  write_iops: 94180
+  write_bandwidth: 505684885
+i7ie.2xlarge:
+  read_iops: 117257
+  read_bandwidth: 1148572714
+  write_iops: 94180
+  write_bandwidth: 505684885
+i7ie.ALL:
+  read_iops: 352834
+  read_bandwidth: 3422623232
+  write_iops: 119327
+  write_bandwidth: 1526442410
--- a/lib/scylla_cloud.py
+++ b/lib/scylla_cloud.py
@@ -812,7 +812,7 @@ class aws_instance(cloud_instance):
         return self._type.split(".")[0]
 
     def is_supported_instance_class(self):
-        if self.instance_class() in ['i2', 'i3', 'i3en', 'c5d', 'm5d', 'm5ad', 'r5d', 'z1d', 'c6gd', 'm6gd', 'r6gd', 'x2gd', 'im4gn', 'is4gen', 'i4i', 'i4g']:
+        if self.instance_class() in ['i2', 'i3', 'i3en', 'c5d', 'm5d', 'm5ad', 'r5d', 'z1d', 'c6gd', 'm6gd', 'r6gd', 'x2gd', 'im4gn', 'is4gen', 'i4i', 'i4g', 'i7ie']:
             return True
         return False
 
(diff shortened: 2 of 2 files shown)


yaronkaikov:
@syuu1228 while you doing that, can you consider also implementing it as a dictionary as described - https://github.com/scylladb/scylla-machine-image/issues/544 ?

common/scylla_cloud_io_setup
```diff
@@ -423,6 +423,21 @@ class aws_io_setup(cloud_io_setup):
...
+        elif self.idata.instancetype in ("i7ie.xlarge", "i7ie.2xlarge"):
+            self.disk_properties["read_iops"] = 117257 * nr_disks
+            self.disk_properties["read_bandwidth"] = 1148572714 * nr_disks
+            self.disk_properties["write_iops"] = 94180 * nr_disks
+            self.disk_properties["write_bandwidth"] = 505684885 * nr_disks
+        elif self.idata.instancetype in ("i7ie.3xlarge", "i7ie.6xlarge", "i7ie.12xlarge", "i7ie.18xlarge", "i7ie.24xlarge", "i7ie.48xlarge"):
+            self.disk_properties["read_iops"] = 352834 * nr_disks
+            self.disk_properties["read_bandwidth"] = 3422623232 * nr_disks
+            self.disk_properties["write_iops"] = 119327 * nr_disks
+            self.disk_properties["write_bandwidth"] = 1526442410 * nr_disks
```
avikivity, line 440: Where do these numbers come from?

Please add to patch changelog.

syuu1228:
> @syuu1228 while you doing that, can you consider also implementing it as a dictionary as described - #544 ?

I'm working on #544, but I found that it takes few more time to complete implementation so I send this first.
We can import i7ie io parameters to #544 patch later.

common/scylla_cloud_io_setup
syuu1228: Unlike other clouds like Azure, EC2 does not have IOPS and bandwidth information on their document.
So all preset values are measured by iotune on target instances instead of referenced from the document.
I measured 3 times on each instance types and used average of the result.

Added same description on the changelog.

yaronkaikov:
@syuu1228 please rebase

common/scylla_cloud_io_setup
avikivity: ok

yaronkaikov (approved):

yaronkaikov:
@avikivity Ok from your POV?

avikivity:
yes

common/scylla_cloud_io_setup
roydahan: previously @xemul had some cavets on how we calculate the numbers for i4i.
I want to make sure we're doing it transparently this time.
Let's have it approved from @xemul as well.

roydahan (changes requested):
Marking as blocked until it gets review for iops and BW numbers.

syuu1228:
> @syuu1228 please add the raw output of perftune to the commit message.

Added.

> Also, can you please elaborate on how you ran iotune? Was it on our AMI? raw instance of i7ie? without installing scylla and starting it first? (I'm asking becuase with i4i we noticed differences)

Here's measurement environment details:
 - Measured on i7ie.* instances with latest version of Ubuntu 24.04 LTS AMI
   (We cannot use Scylla AMI since we do want to measure single drive performance)
 - Measured single local SSD w/o RAID0, since we simulate RAID0 performance on
   scylla_cloud_io_setup script from single drive performance
 - Use iotune for the measurement, executed 3 times for each instance size
   and used average of the results
 - Automated measurement by script: https://github.com/syuu1228/ec2_run_script

roydahan:
Thanks for that.
Do we get the same numbers when we run it on Scylla AMI with raid0 on top of all disks?

syuu1228:
> Thanks for that. Do we get the same numbers when we run it on Scylla AMI with raid0 on top of all disks?

These are theoretical values, so the actual performance may be lower than these figures depending on CPU performance and other factors.

Since EC2 has a fixed number of Local SSDs, it is possible to measure the performance on the RAID directly and apply the actual performance to the parameters, but this would require re-measuring a significant number of instance types.

roydahan:
> Since EC2 has a fixed number of Local SSDs, it is possible to measure the performance on the RAID directly and apply the actual performance to the parameters, but this would require re-measuring a significant number of instance types.

I'm not asking to re-measure on all instances, but take 2-3 instance type and measure with iotune on the raid0 with Scylla AMI and provide results.
They don't have to be exact but at least in the same area.

fruch:
> @syuu1228 meanwhile we did some performance tests using 2024.1.0 and running iotune and we got very good results. I think this one can be merged.
> 
> @fruch do you know if the numbers you got when running the iotune, match the numbers we have here?

https://argus.scylladb.com/tests/scylla-cluster-tests/6ba3a400-6486-4058-9505-4da11dc4dbb0
3 nodes from my last run with iotune from 2024.1.0

```
disks:
  - mountpoint: /var/lib/scylla
    read_iops: 352714
    read_bandwidth: 3460549632
    write_iops: 119234
    write_bandwidth: 1528882304

disks:
  - mountpoint: /var/lib/scylla
    read_iops: 352749
    read_bandwidth: 3450985984
    write_iops: 116640
    write_bandwidth: 1529039104

disks:
  - mountpoint: /var/lib/scylla
    read_iops: 352746
    read_bandwidth: 3451094784
    write_iops: 116078
    write_bandwidth: 1529228160
```

vs. figures from this commit

```
i7ie.ALL:
  read_iops: 352834
  read_bandwidth: 3422623232
  write_iops: 119327
  write_bandwidth: 1526442410
```

%1.8 - %0.15 diff in bandwidth, even less in iops

roydahan (approved):

yaronkaikov:
@avikivity @roydahan @mykaul Any objection about backporting this to 2024.2? 

@tzach was asking about it

avikivity:
No objection

common/scylla_cloud_io_setup
mykaul: https://docs.aws.amazon.com/ec2/latest/instancetypes/so.html#so_instance-store has bandwidth numbers.

common/aws_io_params.yaml
```diff
@@ -558,3 +558,23 @@ i4i.metal:
...
+i7ie.2xlarge:
+  read_iops: 117257
+  read_bandwidth: 1148572714
+  write_iops: 94180
+  write_bandwidth: 505684885
+i7ie.ALL:
+  read_iops: 352834
+  read_bandwidth: 3422623232
+  write_iops: 119327
+  write_bandwidth: 1526442410
```
roydahan, line 580: @syuu1228 / @yaronkaikov now I got back to this after the i7i and i8g PRs and I don't understand it.
Is this one a single disk numbers and we multiply it somewhere by the number of disks?

PR #562, opened 2024-12-09
scylla_cloud_io_setup: move aws io preset values to external YAML file

### LINKED ISSUE #544: Refactor aws_io_setup class - specifically, the IO values per instance

syuu1228:
Since we support many instance types on AWS EC2, handling all instance types by "if - elif" conditions is difficult to read and hard to maintain.

We should move out these preset values to the external file.

Note that, in the original parameter definition, " * nr_disks" was applied to some instance types, but the new code applies it to all instance types.

This is because we know that instance types to which does not applied " * nr_disks" have only one disk.

Fixes #544


syuu1228:
Note that this is not included https://github.com/scylladb/scylla-machine-image/pull/561.
I will rebase it if this one merge first.

yaronkaikov:
@syuu1228 did you verify this change? can you provide the details ?

yaronkaikov:
Verified with https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/7/

yaronkaikov:
> Verified with https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/7/

Based on the AMI created, i have started an i4i.4xlarge instance and the io_properties are working:
```
scyllaadm@ip-10-0-1-171:~$ cat /etc/scylla.d/io_properties.yaml 
disks:
- mountpoint: /var/lib/scylla
  read_bandwidth: 3087631104
  read_iops: 385400
  write_bandwidth: 2289281280
  write_iops: 240628
```

PR #564, opened 2024-12-23
dist: fix incorrect license on rpm & deb packages

syuu1228:


syuu1228:
I found that since we switched our license from AGPL-3.0 to Apache at https://github.com/scylladb/scylla-machine-image/commit/e4116aca5c52c69113c7744082481b5a4c9e52e9, we mistakenly not updated dist/debian/debian/copyright.
So it sill AGPL-3.0 until now, we don't need to update dist/debian to switching license.

syuu1228:
Test passed: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/405/

yaronkaikov:
@avikivity @mykaul please have a look

avikivity:
Why to AGPL? Better to Apache license, we don't want to restrict use here.

syuu1228:
I opened this PR because of the request from @yaronkaikov, which was:
> Do you have any idea how to modify the license for scylla-machine-image from AGPL to APACHE ?
> Can you update about the license change for scylla-machine-image?

So I think current patch is doing opposite.
@avikivity We want use Apache-2.0 instead of AGPL-3, right?

Current license is already Apache-2.0 so it's basically nothing to do here.
But I found one exception:

dist/debian/debian/copyright is still says this code is licensed under AGPL-3.
So I guess we should change this file to Apache-2.

avikivity:
Yes let's change the leftover to Apache2

syuu1228:
Fixed incorrect license on .deb and .rpm.

yaronkaikov (approved):

yaronkaikov:
@avikivity is that ok? can we merge it ?

avikivity (approved):

PR #565, opened 2024-12-23
Install cpufrequtils and run scylla_cpuscaling_setup

### LINKED ISSUE #550: Consistent CPU power configuration
I've noticed that CPU power configuration is not consistent in Scylla Cloud, which as far as I know, it's configured at AMI level. This is totally related to `cpufrequtils` and scaling governors usage.

# Scenarios
## Scenario 1: cpufrequtils is installed but using scaling governor is not supported
`cpufrequtils` is installed, but `/sys/devices/system/cpu/cpufreq/policy0/scaling_governor` doesn't exist, causing `cpufrequtils` to exit.

## Scenario 2: Using a scaling governor is supported but cpufrequtils is not installed
`/sys/devices/system/cpu/cpufreq/policy0/scaling_governor` exists but `cpufrequtils` is not installed and/or not configured as expected (as far as I know, the governor used is `performance`).

# Proposed solution
`cpufrequtils` must be always installed if using a scaling governor is possible. This must happen when the instance is bootstrapped.

syuu1228:
It was mistakenly does not configured as expected, fixing. (Probably we broke this when we switched from Ubuntu minimal base image, since it does not have cpufrequtils)

Fixes #550

### DIFF
--- a/common/scylla_image_setup
+++ b/common/scylla_image_setup
@@ -23,9 +23,7 @@ if __name__ == '__main__':
         run(f'/opt/scylladb/scripts/scylla_swap_setup --swap-directory {swap_directory}', shell=True, check=True)
     machine_image_configured = Path('/etc/scylla/machine_image_configured')
     if not machine_image_configured.exists():
-        # On Ubuntu, we configure CPU scaling while AMI building time
-        if is_redhat_variant():
-            run('/opt/scylladb/scripts/scylla_cpuscaling_setup', shell=True, check=True)
+        run('/opt/scylladb/scripts/scylla_cpuscaling_setup', shell=True, check=True)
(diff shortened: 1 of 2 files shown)


syuu1228:
Test passed: https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/next-machine-image/404/

yaronkaikov (approved):

roydahan:
@syuu1228 do you think it requires performance testing or can affect performance?

mykaul:
Ideally, we should have a test to see it's working.
@roydahan - yes, it can have implications. I doubt, as we did some work to ensure the CPU doesn't go to sleep ('C' state), etc., but I'm not sure about turbo for example.

PR #566, opened 2024-12-24
(cloud images): add `environment` default tag

yaronkaikov:
Adding a logic for the `environment` tag for better image search

The default value is set for `debug`

- master tag value will be `daily`
- releases, the tag value will be `candidate` (and we modify it to `production` once we promote)
- debug jobs (everything running under `machine-image` gating, `byo` or `releng-testing`) the value will be debug

There is one special case for private builds that will be handled in a separate PR

This is a prep for https://github.com/scylladb/scylla-pkg/issues/4077

Need to be merge only after https://github.com/scylladb/scylla-pkg/pull/4725

### DIFF
--- a/packer/build_image.sh
+++ b/packer/build_image.sh
@@ -14,6 +14,7 @@ DRY_RUN=false
 DEBUG=false
 BUILD_MODE='release'
 TARGET=
+ENV_TAG="debug"
 
 print_usage() {
     echo "$0 --repo [URL] --target [distribution]"
@@ -26,6 +27,7 @@ print_usage() {
     echo "  [--branch]              Set the release branch for GCE label. Default: master"
     echo "  [--ami-regions]         Set regions to copy the AMI when done building it (including permissions and tags)"
     echo "  [--build-tag]           Jenkins Build tag"
+    echo "  [--env-tag]             Environment tag for our images. default: debug. Valid options: daily(master)|candidate(releases)|production(releases)|private(custom images for customers)"
     echo "  [--build-mode]          Choose which build mode to use for Scylla installation. Default: release. Valid options: release|debug"
     echo "  [--debug]               Build debug image with special prefix for image name. Default: false."
     echo "  [--log-file]            Path for log. Default build/ami.log on current dir. Default: build/packer.log"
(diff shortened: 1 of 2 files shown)


packer/build_image.sh
```diff
@@ -26,6 +30,7 @@ print_usage() {
     echo "  [--branch]              Set the release branch for GCE label. Default: master"
     echo "  [--ami-regions]         Set regions to copy the AMI when done building it (including permissions and tags)"
     echo "  [--build-tag]           Jenkins Build tag"
+    echo "  [--env-tag]             Environment tag for our images. default: daily for master and candidate for releases. Valid options: debug and production"
```
fruch, line 33: how rc and daily builds would be labeled ?

fruch, line 33: Can we make debug the default  ? so if anyone does it manually it would have `candidate`  nor `daily` ?

yaronkaikov: Sure, done

packer/build_image.sh
```diff
@@ -238,6 +245,7 @@ if [ "$BUILD_MODE" = "debug" ]; then
 fi
 if $DEBUG ; then
   IMAGE_NAME="debug-$IMAGE_NAME"
+  ENV_TAG="debug"
```
Annamikhlin, line 248: Why to add it twice? in line 17 and here..

yaronkaikov: leftover , removed

fruch (dismissed):
LGTM

packer/build_image.sh
```diff
@@ -26,6 +27,7 @@ print_usage() {
     echo "  [--branch]              Set the release branch for GCE label. Default: master"
     echo "  [--ami-regions]         Set regions to copy the AMI when done building it (including permissions and tags)"
     echo "  [--build-tag]           Jenkins Build tag"
+    echo "  [--env-tag]             Environment tag for our images. default: daily for master and candidate for releases. Valid options: debug and production"
```
Annamikhlin, line 30: Can you update the param description as well for default: debug

yaronkaikov: done

Annamikhlin (approved):
LGTM

PR #598, opened 2025-01-09
cleanup: remove `Makefile` `k8s` `source`

yaronkaikov:
As the code is no longer in use


yaronkaikov:
@Annamikhlin @fruch please review

source/user_data_v3.rst
fruch, line 1: is that exists somewhere else ?

fruch: i.e. you don't think this repo need a documentation of it's own ?

fruch: As long as you know where the docs are and how to update them

fruch (approved):

PR #613, opened 2025-01-29
dist: support smooth upgrade from enterprise to source available

syuu1228:
dist: support smooth upgrade from enterprise to source availalbe When upgrading for example from `2024.1` to `2025.1` the package name is not identical casuing the upgrade command to fail.

This makes packages upgradable from enterprise.

Related scylladb/scylladb#22420


yaronkaikov (approved):
LGTM

PR #627, opened 2025-02-10
scylla_install_image: disable cloud provider agents

yaronkaikov:
currently, if the BYOA customer is assigned the required AWS Profile on the instance or will deploy scylla in GCP customer will able to fetch metadata about the OS itself, and also be able to execute a command on a target instance, which may lead to loose of control for instances deployed in BYOA.

Fixes: https://github.com/scylladb/scylla-pkg/issues/4883

### Testing
- [x] https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/21/

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -118,14 +118,17 @@ if __name__ == '__main__':
 
         kernel_opt = ''
         grub_variable = 'GRUB_CMDLINE_LINUX_DEFAULT'
+        run('systemctl mask amazon-ssm-agent', shell=True, check=True)
     elif args.target_cloud == 'gce':
         # align with other clouds image
         run('apt-get purge -y rsyslog', shell=True, check=True)
         kernel_opt = ''
         grub_variable = 'GRUB_CMDLINE_LINUX_DEFAULT'
+        run('systemctl mask google-osconfig-agent', shell=True, check=True)
     elif args.target_cloud == 'azure':
         kernel_opt = ' rootdelay=300'
         grub_variable = 'GRUB_CMDLINE_LINUX'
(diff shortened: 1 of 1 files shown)


d-helios (approved):

packer/scylla_install_image
```diff
@@ -118,14 +118,17 @@ if __name__ == '__main__':
...
     elif args.target_cloud == 'gce':
         # align with other clouds image
         run('apt-get purge -y rsyslog', shell=True, check=True)
         kernel_opt = ''
         grub_variable = 'GRUB_CMDLINE_LINUX_DEFAULT'
+        run('systemctl mask google-osconfig-agent', shell=True, check=True)
     elif args.target_cloud == 'azure':
         kernel_opt = ' rootdelay=300'
         grub_variable = 'GRUB_CMDLINE_LINUX'
+        run('systemctl mask walinuxagent', shell=True, check=True)
```
fruch, line 131: @soyacz 

Aren't we using this for something in SCT ?

soyacz: yes, we use it for rebooting with running OS command (`reboot -f`) due issue with SDK (https://github.com/scylladb/scylla-cluster-tests/pull/6628)
We would need to adjust SCT to enable it back when VM is provisioned.

yaronkaikov: @soyacz Can you open an issue or send a fix? we should probably also have it for 2025.1 since this is about to be backported

yaronkaikov: Thanks @soyacz , I assume we can move forward with this PR then ?

lsfreitas (approved):

Annamikhlin (approved):

PR #660, opened 2025-03-06
Use github action instead of curl method to trigger jenkins

lsfreitas:
Improves the way we trigger next-machine-image job using a dedicated github action instead of using curl.

Fixes: scylladb#659


yaronkaikov (approved):

Annamikhlin:
I don't think the commit msg and the PR msg/title exactly correct with the change

Annamikhlin (approved):
LGTM

Issue #682, opened 2025-04-03
Add packages to scylla ami 2025.1.X


d-helios:
please add following packages to the ami - 2025.1.X

- netcat-openbsd

mykaul:
We really need to agree on what package we install: openbsd-netcat or netcat-openbsd ?

d-helios:
> We really need to agree on what package we install: openbsd-netcat or netcat-openbsd ?

@mykaul for us it's really doens't metter 
we created this issue because of that - https://github.com/scylladb/siren/issues/13101

both netcat-traditional and netcat-openbsd will work for us

d-helios:
> We really need to agree on what package we install: openbsd-netcat or netcat-openbsd ?

thanks, I adjusted issue description

mykaul:
@yaronkaikov - are you backporting it? Did you end up removing ncat?

yaronkaikov:
> [@yaronkaikov](https://github.com/yaronkaikov) - are you backporting it? Did you end up removing ncat?

I didn't remove it, yes i will backport it once it's promoted

PR #683, opened 2025-04-03
images: add netcat-openbsd package

yaronkaikov:
As it required by the cloud team

Fixes: https://github.com/scylladb/scylla-machine-image/issues/682

### Testing
- [ ] https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/30/

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -79,7 +79,7 @@ if __name__ == '__main__':
     run('apt-get update --allow-insecure-repositories -y', shell=True, check=True)
     run('apt-get full-upgrade -y', shell=True, check=True)
     run('apt-get purge -y apport python3-apport fuse', shell=True, check=True)
-    run('apt-get install -y systemd-coredump vim.tiny nmap ncat tmux jq python3-boto xfsprogs mdadm initramfs-tools ethtool vim-nox sysstat cpufrequtils nload htop traceroute dnsutils net-tools', shell=True, check=True)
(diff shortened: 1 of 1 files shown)


packer/scylla_install_image
```diff
@@ -79,7 +79,7 @@ if __name__ == '__main__':
     run('apt-get update --allow-insecure-repositories -y', shell=True, check=True)
     run('apt-get full-upgrade -y', shell=True, check=True)
     run('apt-get purge -y apport python3-apport fuse', shell=True, check=True)
-    run('apt-get install -y systemd-coredump vim.tiny nmap ncat tmux jq python3-boto xfsprogs mdadm initramfs-tools ethtool vim-nox sysstat cpufrequtils nload htop traceroute dnsutils net-tools', shell=True, check=True)
+    run('apt-get install -y systemd-coredump vim.tiny nmap ncat tmux jq python3-boto xfsprogs mdadm initramfs-tools ethtool vim-nox sysstat cpufrequtils nload htop traceroute dnsutils net-tools netcat-openbsd', shell=True, check=True)
```
mykaul, line 82:
```suggestion
    run('apt-get install -y systemd-coredump vim.tiny nmap tmux jq python3-boto xfsprogs mdadm initramfs-tools ethtool vim-nox sysstat cpufrequtils nload htop traceroute dnsutils net-tools netcat-openbsd', shell=True, check=True)
```

mykaul: Wait. Now that I think about it - if we do have ncat already, why do we need netcat? @d-helios ?

yaronkaikov: @d-helios @mykaul do you want me to proceed with that?

mykaul: I'd like to understand it better, but for now, please do. We'll get rid of netcat if we don't need it in the future.

Annamikhlin (approved):

PR #686, opened 2025-04-07
Auto backport - promoted-to-master label not having effect

lsfreitas:
Currently adding a label backport-x.x is enough to open a backport PR.

The issue is that the github action is only checking for backport labels while the script if we have args.pull_request will ignore completely this label again.

With this fix I'm setting the check for labels including the promoted_label outside of the else statement. This way the script will always check it.

Refs: [scylladb#4979](https://github.com/scylladb/scylla-pkg/issues/4979)

### DIFF
--- a/.github/scripts/auto-backport.py
+++ b/.github/scripts/auto-backport.py
@@ -146,17 +146,17 @@ def main():
         closed_prs = [pr]
 
     for pr in closed_prs:
+        labels = [label.name for label in pr.labels]
         if args.pull_request:
             backport_labels = [args.label]
         else:
-            labels = [label.name for label in pr.labels]
             backport_labels = [label for label in labels if backport_label_pattern.match(label)]
-            if promoted_label not in labels:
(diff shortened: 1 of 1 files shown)


Annamikhlin:
Please update the `Refs: scylladb#4979` in the msg.. cause the issue is not in the same repo you need to put the full path

.github/scripts/auto-backport.py
```diff
@@ -151,12 +151,12 @@ def main():
         else:
             labels = [label.name for label in pr.labels]
```
Annamikhlin, line 152: what about to moving it also here? as you did in https://github.com/scylladb/scylla-pkg/pull/5037/?

lsfreitas: Yes yes, you were faster than me 😄 Done! I've also updated the commit message and description for the issue reference

yaronkaikov (approved):

Issue #724, opened 2025-06-10
[GCP] Add support for the new z3 instances


gmizrahi:
GCP is adding smaller instances to the z3 family; we would like to support all.
Google's [page](https://cloud.google.com/compute/docs/storage-optimized-machines) doesn't show all of those new instances, you can follow the table below -

![Image](https://github.com/user-attachments/assets/7cae527b-c46d-45bf-83f7-20c5b84616d0)

Migrated to Jira: [SMI-209](https://scylladb.atlassian.net/browse/SMI-209)

[SMI-209]: https://scylladb.atlassian.net/browse/SMI-209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mykaul:
> the iotune issue: [scylladb/seastar#2802](https://github.com/scylladb/seastar/issues/2802)

This is unrelated to z3, or is it?

mykaul:
> It is related to Z3, we were asked to add support for it

@yaronkaikov - but do we have unreliable/low results also for Z3?

mixellent:
@syuu1228 I found this for the z3 equivalent: https://cloud.google.com/compute/docs/disks/local-ssd#titanium-ssd-perf-nvme
(z3 ssd are different than those for n2)

I see there's data only for the currently available large instance types:
<img width="840" alt="Image" src="https://github.com/user-attachments/assets/88332bef-7e92-4759-b170-92a99867384f" />
Can you infer from it the numbers for smaller instances, or do you need me to ask the GCP team for specific numbers?

mixellent:
@syuu1228 can you please provide an update on this from your end?  Anything you need?
We want to have a joint launch with GCP for these smaller z3 instances on June 18, and so this is a blocker from adding the z3 images to Scylla Cloud.

mixellent:
@yaronkaikov GCP just provided me with the perf results for the z3 instances:

![image](https://github.com/user-attachments/assets/84d608c2-39ec-4585-a283-2cabf78d3a85)

roydahan:
@mixellent at the moment we don't have io-prperties defined for any GCP instance.
It means that every time we launch a new instance, iotune runs and measure the disks capabilities (takes several minutes).
Do we want to add io-properties support for GCP instances?

roydahan:
[...]

PR #728, opened 2025-06-16
[GCP] Add support for the new z3 instances

yaronkaikov:
GCP is adding smaller instances to the z3 family; we would like to support all for Cloud

Fixes: https://github.com/scylladb/scylla-machine-image/issues/724

### Testing 
- [x] https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/32/

### DIFF
--- a/lib/scylla_cloud.py
+++ b/lib/scylla_cloud.py
@@ -358,7 +358,7 @@ class gcp_instance(cloud_instance):
         """Returns if this instance type belongs to supported ones for nvmes"""
         if self.instancetype == self.m1supported:
             return True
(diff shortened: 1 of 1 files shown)


yaronkaikov:
@mixellent to which releases we need to backport this?

fruch:
@soyacz can you share io_properties.yaml from runs you did ? as reference?

soyacz:
for 16 cpu, highlssd:
```
mountpoint: /var/lib/scylla
read_iops: 1574862
read_bandwidth: 6448667136
write_iops: 1086687
write_bandwidth: 5370057216
```
Probably need to measure for all sizes and also standardlssd version

PR #731, opened 2025-06-20
Set Interactive Session Timeout

### LINKED ISSUE #705: CIS compliance: Set Interactive Session Timeout
We need to configure TMOUT parameter on /etc/profile.

Detail description is at:
https://github.com/user-attachments/files/16872980/usg-report-20240420.1112.zip

Rule ID:
- xccdf_org.ssgproject.content_rule_accounts_tmout

syuu1228:
Setting the TMOUT option in /etc/profile ensures that all user sessions will terminate based on inactivity. The value of TMOUT should be exported and read only. The TMOUT setting in a file loaded by /etc/profile, e.g. /etc/profile.d/tmout.sh should read as follows:

```
TMOUT=900
readonly TMOUT export TMOUT
```

Rationale:
Terminating an idle session within a short time period reduces the window of opportunity for unauthorized personnel to take control of a management session enabled on the console or console port that has been left unattended.

 - xccdf_org.ssgproject.content_rule_accounts_tmout

fixes #705
Related scylladb/scylla-pkg#2953

### DIFF
--- a/packer/apply_cis_rules
+++ b/packer/apply_cis_rules
@@ -238,3 +238,12 @@ MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512,h
     with open('/etc/ssh/sshd_config.d/99-cis-rules.conf', 'w') as f:
         f.write(sshd_config)
     run('systemctl restart ssh.service', shell=True, check=True)
+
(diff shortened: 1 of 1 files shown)


yaronkaikov:
I didn't see that this PR was based against master (instead of next), fixed it now

Issue #703, opened 2025-05-07
CIS compliance: Ensure All Users/Groups have Unique UID/GID


syuu1228:
All users/groups should have unique UID/GID.

Detail description is at:
https://github.com/user-attachments/files/16872980/usg-report-20240420.1112.zip

Rule ID:
- xccdf_org.ssgproject.content_rule_account_unique_id
- xccdf_org.ssgproject.content_rule_group_unique_id

Migrated to Jira: [SMI-201](https://scylladb.atlassian.net/browse/SMI-201)

[SMI-201]: https://scylladb.atlassian.net/browse/SMI-201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #734, opened 2025-06-23
Set Password Maximum Age

syuu1228:
To specify password maximum age for new accounts, edit the file /etc/login.defs and add or correct the following line:

PASS_MAX_DAYS 365
A value of 180 days is sufficient for many environments. The DoD requirement is 60. The profile requirement is 365.

Rationale:
Any password, no matter how complex, can eventually be cracked. Therefore, passwords need to be changed periodically. If the operating system does not limit the lifetime of passwords and force users to change their passwords, there is the risk that the operating system passwords could be compromised.

Setting the password maximum age ensures users are required to periodically change their passwords. Requiring shorter password lifetimes increases the risk of users writing down the password in a convenient location subject to physical compromise.

fixes: [SMI-198](https://scylladb.atlassian.net/browse/SMI-198)

[SMI-198]: https://scylladb.atlassian.net/browse/SMI-198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packer/apply_cis_rules
+++ b/packer/apply_cis_rules
@@ -358,3 +358,26 @@ ucredit = -1
     # Write the modified content back to the file
     with open(PAM_CONF, 'w') as f:
         f.write(content)
+
+
+    # xccdf_org.ssgproject.content_rule_accounts_maximum_age_login_defs
+    var_accounts_maximum_age_login_defs = '365'
+
+    # Check if PASS_MAX_DAYS exists in /etc/login.defs
+    with open('/etc/login.defs', 'r') as f:
+        content = f.read()
+
+    if re.search(r'^PASS_MAX_DAYS', content, re.MULTILINE):
+        # Replace existing PASS_MAX_DAYS line
+        new_content = re.sub(r'PASS_MAX_DAYS.*',
+                           f'PASS_MAX_DAYS     {var_accounts_maximum_age_login_defs}',
+                           content)
+        with open('/etc/login.defs', 'w') as f:
+            f.write(new_content)
+        success = True
+    else:
+        success = False
+
+    if not success:
+        with open('/etc/login.defs', 'a') as f:
+            f.write(f"PASS_MAX_DAYS      {var_accounts_maximum_age_login_defs}\n")


fruch:
Why not disable password access completely?

Isn't that considered safer ?

syuu1228:
> Why not disable password access completely?
> 
> Isn't that considered safer ?

Since the issue is based on CIS compliance benchmarks (https://github.com/scylladb/scylla-pkg/issues/2953), it sometimes not really matched our machine-image use cases (ex: we don't usually use password to use the image).
Maybe we can just ignore such items if we just secure our machine image, but if we want to make our image to pass CIS compliance benchmarks I think we need to apply the rule.
On this case, the new settings will just not used for normal usage but will apply once we set the password on the instance.

However, I think some of the issues are even conflict with current machine-image implementation, for example this one: https://github.com/scylladb/scylla-machine-image/issues/703
It means we have to drop "ubuntu" and "centos" user we currently use it for alias as "scyllaadm" user.

And this one (not opened issue yet but marked as "FAILED" on CIS benchmark), it may not safe for machine-image since it is recommended to disabling inactive account, sounds like it may causes unable to login to the instance:
スクリーンショット 2025-06-23 173224

@yaronkaikov Do you have any idea with this?

mykaul:
I'm surprised this is still needed. Please re-verify. From https://specopssoft.com/blog/cis-password-policy/ :
```
Password expiration 
Similar to the [NIST Password Guidance](https://specopssoft.com/blog/nist-password-reset-guidelines/), the CIS believes that periodic password changes are more harmful than beneficial. They offer no containment benefits and enforce bad habits—since they encourage users to choose variants of older passwords. In an effort to scale back, the CIS now recommends an annual password reset. Users inevitably share credentials between accounts, and this measure causes minimal burden. 

Admins should also enforce changes when a password compromised, the user role has changed, or if the user leaves the organization.
```

syuu1228:
Dropped Password Minimum Age since it is not listed on latest CIS benchmark result.

syuu1228:
Rebased with latest next, tested with next-machine-image and longevity:

next-machine-image passed: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/next-machine-image/519/
longevity-100gb-4h-test passed: https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/longevity/job/longevity-100gb-4h-test/65/

syuu1228:
Scylla Cloud test passed (except Monitor error):
https://jenkins.scylladb.com/view/siren/job/siren-jobs/job/siren-backend-e2e-manual/8248/

yaronkaikov (approved):

syuu1228:
@yaronkaikov rebased.

syuu1228:
@yaronkaikov rebased

PR #735, opened 2025-06-24
Red Hat build - move to UBI 10 minimal

mykaul:
Move from Rock 8 to UBI 10, minimal image.

TODO: check if we need to use microdnf vs. dnf
I believe a hack with symbolic link can do the trick.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,8 +26,8 @@ jobs:
             pip install -r test-requirements.txt
             pytest ./tests
 
-      - name: Build RPM (Rockylinux:8)
-        run: docker run -v `pwd`:/scylla-machine-image -w /scylla-machine-image  --rm rockylinux:8 bash -c 'dnf update -y; dnf install -y git ; git config --global --add safe.directory "*"; ./dist/redhat/build_rpm.sh -t centos8'
+      - name: Build RPM (CentOS 8)
+        run: docker run -v `pwd`:/scylla-machine-image -w /scylla-machine-image  --rm redhat/ubi10-minimal bash -c 'ln -s /usr/bin/microdnf /usr/bin/dnf && ln -s /usr/bin/microdnf /usr/bin/yum && dnf update -y && dnf install -y git sudo && git config --global --add safe.directory "*"; ./dist/redhat/build_rpm.sh -t centos8'
 
       - name: Build DEB (Ubuntu:22.04)
         run: docker run -v `pwd`:/scylla-machine-image -w /scylla-machine-image  --rm ubuntu:22.04 bash -c 'apt update -y; apt install -y git ; git config --global --add safe.directory "*"; ./dist/debian/build_deb.sh'


.github/workflows/build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             pytest ./tests
 
       - name: Build RPM (Rockylinux:8)
```
fruch, line 29: Change the title

.github/workflows/build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
...
-        run: docker run -v `pwd`:/scylla-machine-image -w /scylla-machine-image  --rm rockylinux:8 bash -c 'dnf update -y; dnf install -y git ; git config --global --add safe.directory "*"; ./dist/redhat/build_rpm.sh -t centos8'
+        run: docker run -v `pwd`:/scylla-machine-image -w /scylla-machine-image  --rm redhat/ubi10-minimal bash -c 'dnf update -y; dnf install -y git ; git config --global --add safe.directory "*"; ./dist/redhat/build_rpm.sh -t centos8'
```
fruch, line 30: You'll need to use microdnf

fruch, line 30: Also not sure what the -t is used for, probably need to change that.

mykaul: That will break follow up invocations of dnf. I need to symlink microdnf to dnf.

mykaul:
Draft - until I understand why I'm building with UBI10, but a CentOS 8 image? Huh?

.github/workflows/build.yml
mykaul: @yaronkaikov - do you wish me to overhaul dist/redhat/build_rpm.sh to use microdnf and UBI10? I'd be glad to do it.

.github/workflows/build.yml
```diff
@@ -26,8 +26,8 @@ jobs:
             pip install -r test-requirements.txt
             pytest ./tests
 
-      - name: Build RPM (Rockylinux:8)
-        run: docker run -v `pwd`:/scylla-machine-image -w /scylla-machine-image  --rm rockylinux:8 bash -c 'dnf update -y; dnf install -y git ; git config --global --add safe.directory "*"; ./dist/redhat/build_rpm.sh -t centos8'
+      - name: Build RPM (CentOS 8)
```
yaronkaikov, line 29: Just a general question, this is our unit-test, why to invest time moving to ubi? why does it need to be change at all?

fruch: anyhow at the end the actual build would be with dbuild based on fedora

if we want to aligned anything here, it should be with the fedora version within dbuild

not sure how UBI is related here.

yaronkaikov: yes, so unless we move dbuild (and any other containers) to be based on `ubi` i don't think we should change this. We can however use updated OS (like Rockylinux:9)

yaronkaikov: We are just verifying that nothing is breaking our package build process

mykaul: But we are not using our build process, but another process, that we have implemented in this script... 
Confusing.

fruch: Since the build process of this one is spread across multiple repos

This was introduced at a unit test of the part that exists in this repository.

We are trying to tell you that from the start...

PR #746, opened 2025-07-02
aws: add i7i to supported instance family

yaronkaikov:
And set the predefined io_properties only upto `2xlarge` instances


roydahan:
For now, we decided to backport only to 2025.2 which is the next version to be installed for new clusters in scylla-cloud.
Once we have a fix for iotune, we will merge and backport the fixed io.properties numbers to 2025.1 and 2025.3

PR #753, opened 2025-07-07
[RFC] scylla_configure: Add helper script to estimate stream_io_throughput_mb_per_sec (on EC2)

### LINKED ISSUE #742: Update stream_io_throughput_mb_per_sec based on instance type
According to Scylla [recommendation](https://github.com/scylladb/scylladb/issues/24758), `stream_io_throughput_mb_per_sec` should be given a reasonable value.
This issue is to make sure we set the value correctly.

The suggested path:
- Add additional line with the key `stream_io_throughput_mb_per_sec` and the value either based on calculations, instance declared network properties or just a rough estimation based on the instance size low(=0.5GB/s), meduim(=0.75GB/s) and large(=1GB/s)
- in [cloud-setup](https://github.com/scylladb/scylla-machine-image/blob/next/common/scylla_cloud_io_setup) read the value and write it to `sylla.yaml`

Migrated to Jira: [SMI-217](https://scylladb.atlassian.net/browse/SMI-217)

[SMI-217]: https://scylladb.atlassian.net/browse/SMI-217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

elcallio:
Fixes #742

Refs #24419
Refs #24758

Small, terrible script to guestimate an appropriate bandwidth for the above scylla option. This is defined as min(min disk bw, .75 * net bw).

Of course, net bandwidth is almost impossible to estimate, esp. since destination matters. So we cheat wildly and just check the capacity of the network adapter. And not really that even. We check if we are running EC2. If so, we check the current instance type against a list of known capacities for EC2 (generated via awscli).

This may be improved by
a.) Do proper measurement. Problem is against what, since nw bw is destination dependent b.) Add numbers/measurements for other clouds.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,4 +4,4 @@ psutil==7.0.0
 pytest==8.4.1
 PyYAML==6.0.2
 traceback-with-variables==2.2.0
-boto3==1.39.6
+boto3==1.39.14


lib/param_estimation.py
```diff
@@ -0,0 +1,30 @@
...
+            if len(instance_info) != 0:
+                net_bw = int(instance_info[0][2] * 1024 * 1024 * 1024) #GBPS
+        if net_bw != 0:
+            # todo. duplicated from io_setup
+            with open('/opt/scylladb/scylla-machine-image/aws_io_params.yaml') as f:
+                io_params = yaml.safe_load(f)
+                t = cloud_instance.instance_class() + '.ALL'
+                if instance_type in io_params:
+                    t = instance_type
+                disk_bw = min(io_params[t]["read_bandwidth"], io_params[t]["write_bandwidth"])
```
regevran, line 23: Can you explain why do we care about `writes` when it comes to `estimating_streaming_bandwidth`?

elcallio: I assume that streaming not only reads. We are expecting to possibly be on the receiving end, no?

regevran: Let's discuss it in [#24793](https://github.com/scylladb/scylladb/pull/24793) - as it the PR that sets the recommendation this PR follows.

common/scylla_configure.py
```diff
@@ -78,6 +78,9 @@ def updated_ami_conf_defaults(self):
         self.CONF_DEFAULTS["scylla_yaml"]["broadcast_rpc_address"] = private_ip
         self.CONF_DEFAULTS["scylla_yaml"]["seed_provider"][0]['parameters'][0]['seeds'] = private_ip
         self.CONF_DEFAULTS["scylla_yaml"]["endpoint_snitch"] = self.cloud_instance.endpoint_snitch
+        stream_bandw = estimate_streaming_bandwidth()
+        if stream_bandw != 0:
+            self.CONF_DEFAULTS["scylla_yaml"]["stream_io_throughput_mb_per_sec"] = stream_bandw
```
regevran, line 83:
```suggestion
self.CONF_DEFAULTS["scylla_yaml"]["stream_io_throughput_mb_per_sec"] = estimate_streaming_bandwidth()
```
No need to special case zero as it is the default anyhow and means no limit.

lib/param_estimation.py
```diff
@@ -0,0 +1,21 @@
...
+        instance_type = cloud_instance.instancetype
+        with open('/opt/scylladb/scylla-machine-image/aws_net_params.json') as f:
+            netinfo = json.load(f)
+            instance_info = [info for info in netinfo if info[0] == instance_type]
+            if len(instance_info) != 0:
+                net_bw = int(instance_info[0][2] * 1024 * 1024 * 1024) #GBPS
+    # TODO: other clouds
+
+    if net_bw == 0:
+        return 0
```
regevran, line 18:
```suggestion
```
The formula below results with the same value (zero) when `net_be == 0`.
No need a special case here.

regevran (approved):

dist/redhat/scylla-machine-image.spec
```diff
@@ -36,6 +36,7 @@ install -d -m755 $RPM_BUILD_ROOT/opt/scylladb/scylla-machine-image
 install -d -m755 $RPM_BUILD_ROOT/opt/scylladb/scylla-machine-image/lib
 install -m644 lib/log.py lib/scylla_cloud.py lib/user_data.py $RPM_BUILD_ROOT/opt/scylladb/scylla-machine-image/lib
 install -m644 common/aws_io_params.yaml $RPM_BUILD_ROOT/opt/scylladb/scylla-machine-image/
+install -m644 common/aws_net_params.json $RPM_BUILD_ROOT/opt/scylladb/scylla-machine-image/
```
mykaul, line 39: I don't fully understand why we need yet another YAML, only this time it is a JSON file, which has a different set of EC2 instance types. Now when we wish to edit/add instance, we need to different files edited, in 2 different formats.
Just add a single parameter to aws_io_params.yaml - base network throughput.

common/scylla_configure.py
elcallio: But setting it implies we've figured out a reasonable value. Which we have not. I prefer to leave it up to scylla to decide, should default change from 0 in the future.

dist/redhat/scylla-machine-image.spec
elcallio: The file in question can be auto generated by a simple call to aws cli (which I forgot to include a comment about in this PR). It can only output json. The point of using the generated file, and not either manually create it or create a formatting routine is that it remains trivial to generate using existing, non-scylla tools.

lib/param_estimation.py
elcallio: True. Removed disk part, so... Done

common/scylla_configure.py
regevran: The confusion is in the gap between `estimate_streaming_bandwidth`'s API and the the configuration is being used for.
In the function it means "couldn't figure out" .
In the configuration it is "no limit".

But fine, leave it with the `if`, giving a chance to "translate" the function's return to something else.

dist/redhat/scylla-machine-image.spec
regevran: Maybe when we'll have enough outputs from various tools, like the one in question and maybe similar for GCP, Azure, etc. Maybe when these pre-created configuration files are piled up - we'll add yet another script to combine them together into a one coherent source of truth.
I don't see the urgency now.

regevran (approved):

common/aws_net_params.json
```diff
@@ -0,0 +1,4627 @@
+[
+    [
+        "m6id.xlarge",
```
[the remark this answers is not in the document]
mykaul: AI can help with that.

bhalevy:
@regevran please summarize what's required to push this PR forward

regevran:
> @regevran please summarize what's required to push this PR forward

Once I will have results running this [test](https://github.com/regevran/yaron-sct/tree/for_local_cli_scylla_based) we'll be able to proceed.

The results should show that reducing io-stream to 75% from network bandwidth indeed results in good user workload latencies.

I'll update the results here once I have them.

regevran:
First results are [published](https://github.com/scylladb/scylladb/pull/24793#issuecomment-3084349060)

yaronkaikov (approved):

PR #774, opened 2025-08-04
Capture exception stderr from scylla_io_setup child

robertbindar:
During latest effort to get io-properties on i7i family we added scylladb/ PR#25315 and during development of that feature we got some failures related to iotune and it turned out the error logs don't get propagated up to surface because this subprocess launch of scylla_io_tune is called with check=True, capture_output=True and the exception isn't caught to print stderr.


roydahan (approved):

mykaul:
It needs backporting, no?

PR #794, opened 2025-09-09
Set Account Expiration Following Inactivity

syuu1228:
To specify the number of days after a password expires (which signifies inactivity) until an account is permanently disabled, add or correct the following line in /etc/default/useradd:
```
INACTIVE=45
```
If a password is currently on the verge of expiration, then 45 day(s) remain(s) until the account is automatically disabled. However, if the password will not expire for another 60 days, then 60 days plus 45 day(s) could elapse until the account would be automatically disabled. See the useradd man page for more information.

Rationale:
Inactive identifiers pose a risk to systems and applications because attackers may exploit an inactive identifier and potentially obtain undetected access to the system. Disabling inactive accounts ensures that accounts which may not have been responsibly removed are not available to attackers who may have compromised their credentials. Owners of inactive accounts will not notice if unauthorized access to their user account has been obtained.

 - xccdf_org.ssgproject.content_rule_account_disable_post_pw_expiration

Fixes SMI-197
Related scylladb/scylla-pkg#2953

### DIFF
--- a/packer/apply_cis_rules
+++ b/packer/apply_cis_rules
@@ -381,3 +381,11 @@ ucredit = -1
     if not success:
         with open('/etc/login.defs', 'a') as f:
             f.write(f"PASS_MAX_DAYS      {var_accounts_maximum_age_login_defs}\n")
+
+
+    # xccdf_org.ssgproject.content_rule_account_disable_post_pw_expiration
+    var_account_disable_post_pw_expiration = '45'
+    config_file = "/etc/default/useradd"
+    with open(config_file, 'a') as f:
+        f.write('\n')
+        f.write(f"INACTIVE={var_account_disable_post_pw_expiration}\n")


yaronkaikov:
@syuu1228 is there a need to test it through cloud ?

syuu1228:
@yaronkaikov Cloud test passed, please merge this.

yaronkaikov:
> > Before merging, please consider what it means for cloud users who don't usually or never connect their machines.
> 
> This patch disables accounts 45 days after the user's password expires; it does not disable accounts that are not regularly logged into.
> 
> Also, since machine-image user accounts are typically operated without passwords, password expiration itself is unlikely to occur.
> 
> Additionally, since it only changes the default value of the useradd command, it will not apply unless new users are added.

@roydahan I think it is safe to merge now

yaronkaikov:
> So basically, IOW this setting doesn't affect anything, right?

I believe the answer is yes, @syuu1228 please confirm

PR #858, opened 2026-01-04
fix(aws): Make cloud-init wait for the network device

fruch:
On some AWS instances, a race condition can occur where cloud-init
starts before the ENA network interface is fully initialized. This can
cause it to fail when it tries to fetch metadata or perform other
network-dependent tasks.

This commit addresses the issue by introducing a systemd override that
forces the `cloud-init-local` service to wait until the `eth0` device is
available. This ensures the network is ready before cloud-init runs.

Additionally, the `ena` kernel module is now preloaded in initramfs to
ensure it's available as early as possible during the boot process.

Closes: SMI-243

### Testing:

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ami-test/50/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ami-test/51/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ami-test/52/
- [x] 🟢 - i8g.48xlarge  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ami-arm-test/13/

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -12,6 +12,7 @@ import re
 import shutil
 import sys
 from subprocess import run
+from textwrap import dedent
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


avikivity:
We should not make the image worse. We should be able to start in an ipv6-only environment.

fruch:
@soyacz 

I remember you might have seen something like this also in i8g family
can you point to the exact ones, and I'll build image for testing

soyacz:
> > @soyacz
> > I remember you might have seen something like this also in i8g family can you point to the exact ones, and I'll build image for testing
> 
> this is seem to solve the issue for the i8g.48xlarge as well

yes, this is one that was failing

packer/scylla_install_image
```diff
@@ -229,6 +230,29 @@ WantedBy=multi-user.target
             ["sudo", "-u", "ubuntu", "/opt/scylladb/scylla-machine-image/download_authorized_keys"],
         ]
 
+        # Enable ENA driver in initramfs
+        run("echo 'ena' | tee -a /etc/initramfs-tools/modules", shell=True, check=True)
+        run("update-initramfs -u", shell=True, check=True)
+
+        # Create systemd override to wait for eth0 before cloud-init-local
+        # so that cloud-init won't fail reading from metadata service
```
soyacz: if so, maybe better just to wait for metadata service available instead of waiting for eth0 (which name could change, or maybe if we spinned up with multiple NIC's)
`ExecStartPre=/bin/sh -c 'for i in $(seq 1 20); do curl -s --connect-timeout 1 http://169.254.169.254 >/dev/null 2>&1 && exit 0; sleep
  0.5; done; exit 1'`

fruch: > if so, maybe better just to wait for metadata service available instead of waiting for eth0 (which name could change, or maybe if we spinned up with multiple NIC's) `ExecStartPre=/bin/sh -c 'for i in $(seq 1 20); do curl -s --connect-timeout 1 http://169.254.169.254 >/dev/null 2>&1 && exit 0; sleep 0.5; done; exit 1'`

name is fixed, cause of flags we are passing down to the kernel

what you suggested is equivalent to what cloud-init is doing itself, and apply retires there doesn't see to be helping out for this issue (I've tried change the retries on cloud-init, and it didn't helped)

soyacz (approved):
LGTM

PR #947, opened 2026-05-05
scylla_install_image: add apt lock timeout to prevent intermittent build failure

Annamikhlin:
image build failed with error:

    E: Could not get lock /var/lib/apt/lists/lock. It is held by process 1888 (apt)
    E: Unable to lock directory /var/lib/apt/lists/
    subprocess.CalledProcessError: Command 'apt-get update -y'
    returned non-zero exit status 100.

image builds fail intermittently when background apt processes hold /var/lib/apt/lists/lock at the time scylla_install_image runs apt-get update.

seen on https://jenkins.scylladb.com/job/scylla-master/job/oci-image/35/ and also https://jenkins.scylladb.com/job/scylla-master/job/next-machine-image/819/

Fix route all apt-get calls through an apt_get() wrapper that:
1. Polls /var/lib/apt/lists/lock, /var/lib/dpkg/lock, and /var/lib/dpkg/lock-frontend with fuser, waiting up to 5 minutes for background processes to release them.
2. Passes -o DPkg::Lock::Timeout=300 as a safety net for the dpkg lock, which supports built-in retries.

The apt lists lock has no built-in retry mechanism, so the fuser polling is required to cover it.

Fixes: https://scylladb.atlassian.net/browse/SMI-279

### Verification
- [x] oci-image - https://jenkins.scylladb.com/job/scylla-master/job/oci-image/37/

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -12,6 +12,7 @@ import platform
 import re
 import shutil
 import sys
+import time
 from pathlib import Path
 from subprocess import run
 from textwrap import dedent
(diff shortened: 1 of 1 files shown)


yaronkaikov (dismissed):
LGTM

packer/scylla_install_image
vponomaryov: This one also must be updated

Annamikhlin: Thanks. Updated as well.

fruch (approved):
LGTM

PR #961, opened 2026-06-10
common/azure: fix wrong IOPS for 5 Azure instance sizes

soyacz:
## What

read/write IOPS for `L8s_v3`, `L16s_v3`, `L48s_v3`, `L8aos_v4` and read IOPS for `L64as_v4` were measured with a wrong block size, yielding values ~40x too low (256 KB block artifact for the `*s_v3` trio) or otherwise inconsistent with sibling SKUs. Bandwidth columns were correct.

Re-benchmarked the affected SKUs and replaced the IOPS values (plus `L16s_v3` write_bandwidth, which was also off) with the measured numbers.

## Verification

Post-fix sanity vs sibling SKUs:
- L8s_v3 read 894k vs L8as_v3 620k / L8s_v4 549k ✓
- L16s_v3 1.38M vs L16s_v4 1.10M ✓
- L48s_v3 5.34M vs L48s_v4 3.28M ✓
- L8aos_v4 718k vs L8as_v4 549k, now > L2aos_v4 ✓
- L64as_v4 4.40M vs L64s_v4 4.40M ✓

All benchmark deviations now green except IOPS (expected — that's what we fixed).

fixes SMI-237

### DIFF
--- a/common/azure_io_params.yaml
+++ b/common/azure_io_params.yaml
@@ -36,10 +36,10 @@ Standard_L4s_v4:
     write_bandwidth: 749365312
 Standard_L8aos_v4:
   disks:
-  - read_iops: 162204
-    read_bandwidth: 4401689600
-    write_iops: 62388
-    write_bandwidth: 2061798528
+  - read_iops: 718283
+    read_bandwidth: 4392082432
+    write_iops: 287006
+    write_bandwidth: 1985829120
 Standard_L8as_v3:
   disks:
   - read_iops: 620172
@@ -54,10 +54,10 @@ Standard_L8as_v4:
     write_bandwidth: 1499555072
 Standard_L8s_v3:
   disks:
-  - read_iops: 14245
-    read_bandwidth: 3734856960
-    write_iops: 10017
-    write_bandwidth: 2612137984
+  - read_iops: 894399
+    read_bandwidth: 3740749056
+    write_iops: 607182
+    write_bandwidth: 2494925312
 Standard_L8s_v4:
   disks:
   - read_iops: 549031
@@ -90,10 +90,10 @@ Standard_L16as_v4:
     write_bandwidth: 3000305920
 Standard_L16s_v3:
   disks:
-  - read_iops: 28489
-    read_bandwidth: 7467323392
-    write_iops: 20005
-    write_bandwidth: 5135863296
+  - read_iops: 1376027
+    read_bandwidth: 7458523648
+    write_iops: 891599
+    write_bandwidth: 3713746176
 Standard_L16s_v4:
   disks:
   - read_iops: 1098789
@@ -138,10 +138,10 @@ Standard_L48as_v4:
     write_bandwidth: 9004469248
 Standard_L48s_v3:
   disks:
-  - read_iops: 85440
-    read_bandwidth: 22391074816
-    write_iops: 59385
-    write_bandwidth: 15443830784
+  - read_iops: 5344012
+    read_bandwidth: 22344859648
+    write_iops: 3523259
+    write_bandwidth: 14548556800
 Standard_L48s_v4:
   disks:
   - read_iops: 3280604
@@ -156,10 +156,10 @@ Standard_L64as_v3:
     write_bandwidth: 19616075776
 Standard_L64as_v4:
   disks:
-  - read_iops: 1115399
-    read_bandwidth: 23921129472
-    write_iops: 1759073
-    write_bandwidth: 12010074112
+  - read_iops: 4403160
+    read_bandwidth: 23945021440
+    write_iops: 1740036
+    write_bandwidth: 10790373376
 Standard_L64s_v3:
   disks:
   - read_iops: 5506751


fruch:
@soyacz is there an need to update https://github.com/scylladb/scylla-machine-image/blob/next/tools/io_properties_measurements/get_scylla_io_properties_azure.py ? 

or it's a fix change in scylla iotune itself ?

soyacz:
> @soyacz is there an need to update https://github.com/scylladb/scylla-machine-image/blob/next/tools/io_properties_measurements/get_scylla_io_properties_azure.py ?
> 
> or it's a fix change in scylla iotune itself ?

## Root cause analysis: wrong IOPS values

These numbers aren't measured by our `get_scylla_io_properties_azure.py` script — the script only SSHes into the VM and reads `/etc/scylla.d/io_properties.yaml`, which is generated by ScyllaDB's `iotune` (`scylla_io_setup`). So the bad values originate in iotune, and we may be better off fixing it there.

**The fingerprint:** for all three broken `*s_v3` entries, `read_bandwidth / read_iops == 262144` exactly (= 256 KB), and the same on the write side.

iotune runs two independent passes:
- random **4 KB** reads → `read_iops`
- sequential **256 KB** reads → `read_bandwidth`

Normally `read_iops` (4 KB) is far larger than `read_bandwidth / 256 KB`. In the broken entries `read_iops` equals `read_bandwidth ÷ 256 KB`, i.e. the recorded IOPS is just the IO count from the **bandwidth pass**, not from the random-IOPS pass. The random small-block measurement was invalid and collapsed to the bandwidth-limited result.

### Possible causes (iotune side)
1. iotune ran before the disk subsystem was ready — local NVMe still attaching or the RAID array not fully assembled — so the random-IOPS phase measured a degraded device.
2. iotune's IOPS phase got throttled / had too short a measurement window and saturated on bandwidth instead of resolving IOPS.
3. iotune fell back to bandwidth-derived IOPS instead of failing loudly when the random phase produced no usable result.

### Possible causes / contributing bugs (our harvesting script)
1. `get_wait_times()` scales wait time off the wrong token: `vm_size.split("_")[-1]` yields the version (`v3`/`v4`), not the core-count token (`L8s`). So the `"8"`/`"16"`/`"32"` checks never match and every v3/v4/v5 VM gets a flat 1.5× wait, with no extra warm-up for large SKUs. This likely let some big instances be read too early.
2. Without `--override`, the script just reads the pre-existing `io_properties.yaml` (baked at image first-boot, possibly measured under bad disk topology) — no fresh measurement happens.

### Suggested fixes
- **Preferred:** make iotune fail/retry instead of silently emitting bandwidth-derived IOPS, and/or ensure it runs only after disks/RAID are fully ready.
- In our script: fix `get_wait_times()` to use the core-count token, run with `--override`, and add a sanity guard that rejects any result where `read_iops * 262144 ≈ read_bandwidth` (flag for rerun rather than writing it).

fruch:
> > @soyacz is there an need to update [`next`/tools/io_properties_measurements/get_scylla_io_properties_azure.py](https://github.com/scylladb/scylla-machine-image/blob/next/tools/io_properties_measurements/get_scylla_io_properties_azure.py?rgh-link-date=2026-06-10T11%3A18%3A26Z) ?
> > or it's a fix change in scylla iotune itself ?
> 
> ## Root cause analysis: wrong IOPS values
> These numbers aren't measured by our `get_scylla_io_properties_azure.py` script — the script only SSHes into the VM and reads `/etc/scylla.d/io_properties.yaml`, which is generated by ScyllaDB's `iotune` (`scylla_io_setup`). So the bad values originate in iotune, and we may be better off fixing it there.
> 
> **The fingerprint:** for all three broken `*s_v3` entries, `read_bandwidth / read_iops == 262144` exactly (= 256 KB), and the same on the write side.
> 
> iotune runs two independent passes:
> 
> * random **4 KB** reads → `read_iops`
> * sequential **256 KB** reads → `read_bandwidth`
> 
> Normally `read_iops` (4 KB) is far larger than `read_bandwidth / 256 KB`. In the broken entries `read_iops` equals `read_bandwidth ÷ 256 KB`, i.e. the recorded IOPS is just the IO count from the **bandwidth pass**, not from the random-IOPS pass. The random small-block measurement was invalid and collapsed to the bandwidth-limited result.
> 
> ### Possible causes (iotune side)
> 1. iotune ran before the disk subsystem was ready — local NVMe still attaching or the RAID array not fully assembled — so the random-IOPS phase measured a degraded device.
> 2. iotune's IOPS phase got throttled / had too short a measurement window and saturated on bandwidth instead of resolving IOPS.
> 3. iotune fell back to bandwidth-derived IOPS instead of failing loudly when the random phase produced no usable result.
> 
> ### Possible causes / contributing bugs (our harvesting script)
> 1. `get_wait_times()` scales wait time off the wrong token: `vm_size.split("_")[-1]` yields the version (`v3`/`v4`), not the core-count token (`L8s`). So the `"8"`/`"16"`/`"32"` checks never match and every v3/v4/v5 VM gets a flat 1.5× wait, with no extra warm-up for large SKUs. This likely let some big instances be read too early.
> 2. Without `--override`, the script just reads the pre-existing `io_properties.yaml` (baked at image first-boot, possibly measured under bad disk topology) — no fresh measurement happens.
> 
> ### Suggested fixes
> * **Preferred:** make iotune fail/retry instead of silently emitting bandwidth-derived IOPS, and/or ensure it runs only after disks/RAID are fully ready.
> * In our script: fix `get_wait_times()` to use the core-count token, run with `--override`, and add a sanity guard that rejects any result where `read_iops * 262144 ≈ read_bandwidth` (flag for rerun rather than writing it).

@soyacz can we use the code/script to collect the iotune data, to validate it doing the correct thing ?
after that we can further check if the figures wouldn't match.

fruch:
## Independent iotune validation (scylla-2026.1.5 image, eastus, `--override`)

Re-ran `tools/io_properties_measurements/get_scylla_io_properties_azure.py` with `--override` (forces fresh `scylla_io_setup` / iotune benchmark) on all 5 affected SKUs using the `scylla-2026.1.5` gallery image in eastus.

### Results

| VM Size | Metric | Fresh iotune | PR value | Ratio | Notes |
|---------|--------|-------------|-----------|-------|-------|
| **L8s_v3** | read_iops | 688,549 | 894,399 | 0.77x | 1 NVMe, min_io=4096 |
| | read_bw | 3,720 MB/s | 3,740 MB/s | ~1.0x | |
| | write_iops | 447,904 | 607,182 | 0.74x | |
| | write_bw | 1,848 MB/s | 2,494 MB/s | 0.74x | |
| **L16s_v3** | read_iops | 28,489 | 1,376,027 | **0.02x** | 2 NVMe md0, min_io=**262144** (256KB bug!) |
| | read_bw | 7,466 MB/s | 7,458 MB/s | ~1.0x | bandwidth correct |
| | write_iops | 20,043 | 891,599 | 0.02x | |
| | write_bw | 5,239 MB/s | 3,713 MB/s | 1.4x | |
| **L48s_v3** | read_iops | 5,343,082 | 5,344,012 | ~1.0x | 6 NVMe md0, min_io=4096 |
| | read_bw | 22,355 MB/s | 22,344 MB/s | ~1.0x | |
| | write_iops | 3,530,263 | 3,523,259 | ~1.0x | |
| | write_bw | 14,601 MB/s | 14,548 MB/s | ~1.0x | |
| **L8aos_v4** | read_iops | 719,128 | 718,283 | ~1.0x | 6 NVMe md0, min_io=4096 |
| | read_bw | 4,404 MB/s | 4,392 MB/s | ~1.0x | |
| | write_iops | 287,774 | 287,006 | ~1.0x | |
| | write_bw | 2,202 MB/s | 1,985 MB/s | 1.1x | |
| **L64as_v4** | read_iops | 4,396,940 | 4,403,160 | ~1.0x | 8 NVMe md0, min_io=4096 |
| | read_bw | 23,945 MB/s | 23,945 MB/s | ~1.0x | |
| | write_iops | 1,758,911 | 1,740,036 | ~1.0x | |
| | write_bw | 11,974 MB/s | 10,790 MB/s | 1.1x | |

### Key findings

1. **L48s_v3, L8aos_v4, L64as_v4**: Fresh iotune matches PR values almost exactly (within 1-2% on IOPS, ~10% on write bandwidth). PR values are validated.

2. **L16s_v3**: iotune picks `minimum_io_size=262144` (256KB) on the 2-disk md0 array, reproducing the exact bug described in the PR. The 28k IOPS result is ~48x lower than the correct ~1.3M value. This confirms the PR description — the original wrong values came from iotune using the wrong block size on this specific disk configuration.

3. **L8s_v3**: Single NVMe (no RAID), iotune uses correct 4096 block size. IOPS ~25% lower than PR values — normal single-run variance (PR likely used best-of-multiple or a different measurement run).

### Conclusion

The tool works correctly and reproduces the bug for L16s_v3 (confirming the PR fix is needed). For all other SKUs, fresh measurements match the PR values within expected variance. The tooling is validated for measuring new instance types going forward.

---
*Ran with: `uv run python tools/io_properties_measurements/get_scylla_io_properties_azure.py --override --image-id .../scylladb/images/scylla-2026.1/versions/2026.1.5 --location eastus`*

fruch:
## Cross-reference: upstream iotune bug

The root cause of the wrong IOPS values is a known iotune block-size detection issue tracked across multiple repos:

### Bug chain

| Tracker | Issue | Status | Description |
|---------|-------|--------|-------------|
| **Seastar** | [#1698](https://github.com/scylladb/seastar/issues/1698) | Closed | iotune: Random IO buffer size is not always correct |
| **Seastar** | [#2802](https://github.com/scylladb/seastar/issues/2802) | Closed | iotune: measurement results does not scale on larger nodes of EC2 |
| **Seastar** | [#3046](https://github.com/scylladb/seastar/pull/3046) | Merged (Dec 2025) | file: Query physical block size and minimum I/O size |
| **ScyllaDB** | [#25211](https://github.com/scylladb/scylladb/issues/25211) | Closed (P1, backport/2025.4-done) | Make `scylla_io_setup` aware of iotune block sizes |
| **ScyllaDB** | [#25315](https://github.com/scylladb/scylladb/pull/25315) | Closed (promoted-to-master, backport/2025.4-done) | Make scylla_io_setup detect request size for best IOPS |
| **Jira** | [SMI-237](https://scylladb.atlassian.net/browse/SMI-237) | In Review (P3) | Preconfigured io-properties on Azure image are deviating by 80-120% from measured |

### Finding: upstream fix does NOT resolve 2-disk md arrays

Despite the iotune fix being in 2025.4+ and master (thus present in 2026.1.5 that we tested with), **the L16s_v3 still gets `minimum_io_size=262144` (256KB)** when it has 2 NVMe disks combined into md0.

iotune disk detection per our test:

| VM Size | NVMe disks | Array | min_io_size | IOPS correct? |
|---------|-----------|-------|-------------|---------------|
| L8s_v3 | 1 | none | 4096 | Yes |
| L16s_v3 | 2 | md0 | **262144** | **No** (48x too low) |
| L48s_v3 | 6 | md0 | 4096 | Yes |
| L8aos_v4 | 6 | md0 | 4096 | Yes |
| L64as_v4 | 8 | md0 | 4096 | Yes |

The pattern: **specifically the 2-disk md RAID** triggers wrong block-size detection. Single disks and arrays with 6+ disks work correctly. This suggests the upstream fix in seastar#3046/scylladb#25315 did not fully resolve the md array edge case.

### Conclusion

This PR's approach of pre-configuring the correct io-properties in the image is the right workaround — the upstream iotune fix does not cover all affected configurations (particularly the 2-disk md case on L16s_v3). The pre-baked values bypass iotune's detection entirely and provide known-good numbers.

The iotune regression for 2-disk md arrays may warrant a follow-up issue in seastar/scylladb.

[SMI-237]: https://scylladb.atlassian.net/browse/SMI-237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

soyacz:
@fruch I just reran artifact tests to get these values. Seems to be ok (or at least ok-ish  as for L8s_v3 you have different value. I've tested with recent master).

fruch:
## Update: 2-disk md bug confirmed fixed on master

Re-ran iotune on **L16s_v3** (2 NVMe disks in md0) with latest master dev image (`2026.3.0~dev-0.20260609.b794bf1204a1`):

```
minimum_io_size=4096  ✅ (was 262144 on 2026.1.5)
read_iops:  1,783,787
write_iops: 1,176,368
```

The upstream fix (seastar#3046 / scylladb#25315) **does** cover the 2-disk md array case — it just has not been backported to 2026.1 yet. Pre-configured values in this PR remain the correct workaround for release branches that lack the fix.

fruch (approved):

PR #119, opened 2021-04-05
fix(scylla_configure): missing newline in the scylla yaml auto generated comment

bentsi:

Issue #118, opened 2021-04-05
I can't build unified-deb on 4.5 due to a machine-image issue


hagitsegev:
https://jenkins.scylladb.com/view/scylla-4.5/job/scylla-4.5/job/unified-deb/4/
```
13:19:41  python3-pip-20.2.2-1.fc33.noarch
13:19:41  python3-pip already installed.
13:19:41  Building in /jenkins/slave/workspace/scylla-4.5/unified-deb/scylla-machine-image...
13:19:41  /jenkins/slave/workspace/scylla-4.5/unified-deb/scylla-machine-image
13:19:41  This package has a Debian revision number but there does not seem to be
13:19:41  an appropriate original tar file or .orig directory in the parent directory;
13:19:41  (expected one of scylla-machine-image_4.5~rc0-20210405.0ef5d340d7f.orig.tar.gz, scylla-machine-image_4.5~rc0-20210405.0ef5d340d7f.orig.tar.bz2,
13:19:41  scylla-machine-image_4.5~rc0-20210405.0ef5d340d7f.orig.tar.lzma,  scylla-machine-image_4.5~rc0-20210405.0ef5d340d7f.orig.tar.xz or scylla-machine-image.orig)
13:19:41  continue anyway? (y/n) Use of uninitialized value $ans in pattern match (m//) at /usr/bin/debuild line 1009.
```

fruch:
@syuu1228 do you know of this issue ? as to why it's failing to build a debian package here ?

hagitsegev:
Not sure the issue is in this repo, but opened here, as the sympthom is here.

fruch:
sounds again related to the version string, hence it expect a source tarball, which we don't have:
https://stackoverflow.com/questions/61646468/build-a-debian-package-without-tarball

PR #968, opened 2026-06-13
packer: tolerate cloud-init recoverable errors (exit 2) in Azure build

Annamikhlin:
## Problem

The first inline provisioner in the Azure image build runs:

```sh
sudo /usr/bin/cloud-init status --wait
```

Per the [cloud-init docs](https://docs.cloud-init.io/en/latest/reference/cli.html#status), `cloud-init status` exits:

- `0` — boot finished cleanly
- `1` — cloud-init crashed (a real failure)
- `2` — boot finished but hit **recoverable errors** (a degraded-but-fine state)

Packer's default inline shebang is `/bin/sh -e`, so a `cloud-init status --wait` return of **`2`** aborted the provisioner immediately — before `scylla_install_image` even ran — failing the whole Azure build with `Script exited with non-zero exit status: 2`. Because the degraded boot state is transient, this produced **intermittent** Azure build failures with no useful output (`SMI-278`).

This was misdiagnosed previously as a `scylla_install_image` failure; in reality the script never ran (no apt output appears), and it has no exit-2 path. The failing command is the cloud-init check that runs first.

### Seen in

- https://jenkins.scylladb.com/job/scylla-2026.2/job/next-machine-image/19/ 
- https://jenkins.scylladb.com/job/scylla-master/job/next-machine-image/832/ (master)

## Fix

Capture the exit code and branch on it instead of letting `set -e` treat every non-zero as fatal: `1` fails the build, `2` logs a warning and continues, `0` continues normally. The warning line makes the intermittent path observable in CI logs.

## Verification

Ran the `next-machine-image` job (releng-testing) with `BUILD_AZURE` + `RUN_AZURE_TEST` only (other clouds/tests disabled), against this fork's branches.

| Build | Branch | RELENG | Build & Test Azure Result |
|---|---|---|---|
| [#119](https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/119/) | `fix-azure-cloudinit-exit2-next` | `next` | **SUCCESS** |
| [#117](https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/117/) | `fix-azure-cloudinit-exit2-next` | `next` | **SUCCESS** |
| [#118](https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/118/) | `fix-azure-cloudinit-exit2-2026.2` (backport) | `next-2026.2` | **SUCCESS** |

**Key evidence — the fix works on the real failure path.** Build #116 actually hit the intermittent degraded boot, and the fix let it continue past the exit-2 that previously broke the build:

```
==> azure-arm.azure: WARNING: cloud-init finished with recoverable errors (exit 2); continuing
```

Fixes: SMI-278

### DIFF
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -238,7 +238,7 @@
     },
     {
       "inline": [
-        "sudo /usr/bin/cloud-init status --wait",
+        "cloudinit_status=0; sudo /usr/bin/cloud-init status --wait || cloudinit_status=$?; if [ \"$cloudinit_status\" -eq 1 ]; then echo 'cloud-init failed (exit 1)' >&2; exit 1; elif [ \"$cloudinit_status\" -eq 2 ]; then echo 'WARNING: cloud-init finished with recoverable errors (exit 2); continuing'; fi",
         "sudo /tmp/scylla_install_image --target-cloud {{build_name}} --scylla-version {{user `scylla_full_version`}} {{user `install_args`}}"
       ],
       "type": "shell"


yaronkaikov (approved):
LGTM

fruch:
@Annamikhlin how does it gurrente to fix anything, it just gonna move some unknown failure further down the line

I would rather collect more logs and diagnostics, and not assume everything is right and move on.

Annamikhlin:
> how does it gurrente to fix anything, it just gonna move some unknown failure further down the line
> 
> I would rather collect more logs and diagnostics, and not assume everything is right and move on.

Sure, I checked the cloud-init diagnostics and logs from the failing build before making this change by `cloud-init status --long`.

The build https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/116/
failed because `cloud-init status --wait` returned `rc=2`

The reported "recoverable error" was:
```
extended_status: degraded done
detail: DataSourceAzure [seed=/dev/sr0]
errors: []
recoverable_errors:
  WARNING:
    - Polling IMDS failed attempt 1 with exception: UrlError('404 Client Error: Not Found for url:
      http://169.254.169.254/metadata/reprovisiondata?api-version=2019-06-01')
```
Cloud-init reported `errors: []`, which indicates that provisioning completed successfully and there were no actual cloud-init failures. This is a known recoverable warning rather than a critical system failure.

The warning comes from cloud-init's Azure datasource querying the IMDS `reprovisiondata` endpoint during boot. That endpoint is only relevant for Azure preprovisioned/pooled VMs. Our Packer-created VMs are not preprovisioned, so IMDS returns `404 Not Found`, which cloud-init records as a recoverable warning.

The change not ignoring all cloud-init failures. Exit code `1` still fails the build immediately. The only behavioral change is that exit code `2` (recoverable warnings) no longer causes the build to fail.

fruch (approved):
LGTM, as long those warning are visiable in the build log

PR #978, opened 2026-06-30
fix(create-devices): rescan NVMe controllers when no data disk appears

soyacz:
## Problem

On some OCI and Azure VMs a local **instance-store NVMe controller enumerates but its namespace block device (`nvme0n1`) never appears**. With the controller visible in `dmesg` (`nvme nvme0: pci function ...`) but no namespace node in `/dev`, `scylla_create_devices --data-device instance_store` finds no data disk and aborts:

```
scylla_create_devices ... -> "No device names discovered after N seconds" -> "No data disk found, abort setup"
scylla-image-setup.service: status=1/FAILURE
/etc/scylla/machine_image_configured  -> never written
```

Downstream, the consumer waiting on `machine_image_configured` times out (600s) and node setup fails. Observed repeatedly on OCI (`VM.DenseIO.E5.Flex`) and Azure (`Standard_L8s_v4`); a single VM per cluster is hit while the rest set up fine.

Reported in SCT-447 (OCI) and SCT-297 (Azure).

## Root cause

`wait_for_devices` retries by re-listing `/dev` every 5s. That can **never** surface a namespace the kernel has not probed — re-reading the device list does nothing to make an un-enumerated namespace appear. So the wait window elapses with no effect and setup aborts even though the controller is present.

Re-listing the device node is a no-op for a namespace the kernel hasn't probed; the namespace has to be re-enumerated. The `rescan_controller` / `nvme ns-rescan` poke is the recognized remedy for the "controller present, namespace missing" symptom regardless of why the probe lagged (slow/failed hypervisor-side attach, or a kernel scan bug) — see Related below.

Note on kernel: these images build on Ubuntu 26.04 (OCI/AWS/GCE) and Ubuntu 24.04 LTS (Azure) per `packer/build_image.sh`, i.e. kernels well past the 6.1.x NVMe-scan regression. So 6.1.x is a documented analog of the symptom/remedy, not the diagnosis here; the fix is deliberately cause-agnostic.

## Fix

Between empty retries, trigger a **best-effort kernel re-probe** of NVMe controllers:

- write `1` to each `/sys/class/nvme/nvme*/rescan_controller` (re-scans namespaces; no `nvme-cli` dependency),
- fall back to a PCI bus rescan (`/sys/bus/pci/rescan`) when that sysfs node is absent.

Errors are swallowed — it is a recovery attempt, not a hard requirement, so it stays a no-op on clouds/instances where it does not apply (e.g. attached/`sd*` disks). The existing `device_wait_seconds` window governs how long we keep re-probing.

## Tests

`tests/test_scylla_create_devices.py`:
- rescan is invoked on each empty retry until a device appears,
- rescan writes to every per-controller sysfs node,
- rescan falls back to PCI bus rescan when no per-controller node exists.

All tests pass; `ruff check` / `ruff format` clean.

## Refs

- SCT-447 — `[OCI] NodeSetupFailed: instance-store NVMe data disk never enumerated`
- SCT-297 — same `is_machine_image_configured` timeout on Azure

## Related (public)

The same controller-present/namespace-missing symptom and the `rescan_controller` remedy are documented upstream:

- [platima/nvme-rescan-fix](https://github.com/platima/nvme-rescan-fix) — documents the 6.1.x-kernel NVMe namespace-enumeration regression and ships a systemd boot-time `rescan_controller` service (same mechanism this PR applies in-loop). Durable fix: kernel 6.12+.
- [nvme-cli `ns-rescan` docs](https://github.com/linux-nvme/nvme-cli/blob/master/Documentation/nvme-ns-rescan.txt) — `nvme ns-rescan /dev/nvme0`, the userspace equivalent of the sysfs write.
- [Red Hat: NVMe namespace/ID numbering inconsistent](https://access.redhat.com/solutions/6967589)
- Azure Lsv4 (the `Standard_L8s_v4` case) is fully NVMe with no SCSI fallback; missing local NVMe also occurs when the OS image lacks NVMe support — [Lsv4 series](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/storage-optimized/lsv4-series), [NVMe Temp Disks FAQ](https://learn.microsoft.com/en-us/azure/virtual-machines/enable-nvme-temp-faqs).

Note: this rescan recovers a kernel-failed namespace probe. It cannot help if the image genuinely lacks NVMe drivers (image-build concern), and the durable fix for the kernel regression is shipping images on 6.12+.

### DIFF
--- a/common/scylla_create_devices
+++ b/common/scylla_create_devices
@@ -33,20 +33,46 @@ class NoDiskFoundError(Exception):
         return "No data disk found, abort setup"
 
 
+NVME_PCI_CLASS_PREFIX = "0x0108"  # PCI class: Non-Volatile memory controller
+
+
+def _unbound_nvme_pci_addresses() -> list[str]:
+    """PCI addresses with an NVMe class code that currently have no driver bound.
+
+    A driver can end up detached from a device (e.g. an `unbind` mid-boot, or a
+    probe race) without the device itself leaving the PCI bus. A bus rescan is a
+    no-op in that case: the kernel already has a struct pci_dev for that slot,
+    and pci_scan_slot() only creates new ones for slots it hasn't seen before.
+    """
+    addresses = []
+    for class_file in glob.glob("/sys/bus/pci/devices/*/class"):
+        try:
+            is_nvme = Path(class_file).read_text().strip().startswith(NVME_PCI_CLASS_PREFIX)
+        except OSError:
+            continue
+        if not is_nvme:
+            continue
+        addr = Path(class_file).parent.name
+        if not Path("/sys/bus/pci/devices", addr, "driver").exists():
+            addresses.append(addr)
+    return addresses
+
+
 def rescan_nvme_controllers() -> None:
     """Best-effort: ask the kernel to re-probe NVMe controllers for namespaces.
 
-    On some OCI/Azure VMs the NVMe controller enumerates but its namespace block
-    device (e.g. nvme0n1) never appears, so the local data disk stays invisible
-    and setup aborts with "No data disk found". Writing to each controller's
-    sysfs `rescan_controller` re-scans its namespaces without needing nvme-cli.
-
-    Always also issues a PCI bus rescan, unconditionally: on NVMe-root VMs (e.g.
-    Azure Standard_Lsv4), the root disk has its own healthy, unrelated NVMe
-    controller with a working `rescan_controller` - gating the bus rescan on
-    "no controller responded" would report success from the root's controller
-    and skip the bus rescan entirely, even though a *different*, still-missing
-    local-disk controller is the actual problem the bus rescan could find.
+    Covers three distinct ways a local NVMe data disk can stay invisible on boot,
+    each seen on OCI and/or Azure:
+      - controller enumerated, namespace never appeared: nudge via the
+        controller's own sysfs `rescan_controller`.
+      - controller present on the PCI bus but its driver never got (re)bound:
+        a bus rescan can't fix this - the kernel already knows the slot - so
+        explicitly ask the driver core to bind it via `drivers_probe`.
+      - controller not on the PCI bus at all (yet): a full bus rescan.
+    All three are always attempted, regardless of whether an earlier one
+    reported success: on NVMe-root VMs (e.g. Azure Standard_Lsv4), the root
+    disk has its own healthy, unrelated NVMe controller, so gating a later step
+    on "nothing found yet" would mask a different, still-missing controller.
     Errors are swallowed - this is a recovery attempt, not a hard requirement.
     """
     for ctrl in glob.glob("/sys/class/nvme/nvme*/rescan_controller"):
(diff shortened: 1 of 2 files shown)


soyacz:
@fruch any idea how to test this one?

fruch:
> @fruch any idea how to test this one?

building all 4 image (ignoring ARM for a sec), and running artifact tests, that kind of the minimum
but once you have images, you can give it a go with longevity tests (i.e. more nodes)

I don't know how to directly cause it, cause I have zero clue what's actually happening.

so we need to make sure it's not breaking any of the cloud platform, but since it's on the retry sequence, maybe for testing put it on the top of the flow, to make sure we exercise this code on each image.

regardless, if it might be fixing the issue, I would want this to be working even if `wait_for_devices: 0`  
the fact we set `wait_for_devices: 60` in SCT is cause of a bug we couldn't identify it's root source.

soyacz:
building here: https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/126/

fruch:
@soyacz 

can you confirm if it's was tested with `wait_for_devices: 0`  ?

soyacz:
> @soyacz
> 
> can you confirm if it's was tested with `wait_for_devices: 0` ?

I think we use 60s as a default, aren't we?
How to set it to 0?
Still, [I work on building reproducer](https://github.com/scylladb/scylla-cluster-tests/pull/15372) and test case to see this in action

fruch:
> > @soyacz
> > 
> > can you confirm if it's was tested with `wait_for_devices: 0` ?
> 
> I think we use 60s as a default, aren't we?
> How to set it to 0?
> Still, [I work on building reproducer](https://github.com/scylladb/scylla-cluster-tests/pull/15372) and test case to see this in action

We use as default in SCT 60s, from not that long ago, our users aren't using it.

You can change it in SCT and test

soyacz:
> > > @soyacz
> > > can you confirm if it's was tested with `wait_for_devices: 0` ?
> > 
> > 
> > I think we use 60s as a default, aren't we?
> > How to set it to 0?
> > Still, [I work on building reproducer](https://github.com/scylladb/scylla-cluster-tests/pull/15372) and test case to see this in action
> 
> We use as default in SCT 60s, from not that long ago, our users aren't using it.
> 
> You can change it in SCT and test

I think without it we don't have a chance to run this code.
I've noticed we hit the issue with new Azure instances only (v4) - old ones don't show problems

fruch:
> > > > @soyacz
> > > > can you confirm if it's was tested with `wait_for_devices: 0` ?
> > > 
> > > 
> > > I think we use 60s as a default, aren't we?
> > > How to set it to 0?
> > > Still, [I work on building reproducer](https://github.com/scylladb/scylla-cluster-tests/pull/15372) and test case to see this in action
> > 
> > We use as default in SCT 60s, from not that long ago, our users aren't using it.
> > 
> > You can change it in SCT and test
> 
> I think without it we don't have a chance to run this code.

Exactly my concern, that we fix it like this (assuming it fixing) just if one uses non default configuration.

So one we do have a reproducer, or some suspect fix, I want it to be working out of the box,  not to be dependent on specific non default configuration 

> I've noticed we hit the issue with new Azure instances only (v4) - old ones don't show problems

soyacz:
not reproducing anymore

PR #980, opened 2026-07-01
packer: retry AMI build on InsufficientInstanceCapacity

Annamikhlin:
The AWS amazon-ebs builder pins a single subnet/AZ up-front (scylla.json subnet_filter selects a random "image-build-subnet*") and has no native cross-AZ fallback. When the chosen AZ is out of capacity for the build instance type, packer fails the whole build with no retry.

Seen in https://jenkins.scylladb.com/job/scylla-2026.1/job/next-machine-image/36/ 
aarch64 (Build AMI), which failed the entire pipeline:

```
  ==> amazon-ebs.aws: Error launching source instance: operation error EC2:
  RunInstances, exceeded maximum number of attempts, 10, https response error
  StatusCode: 500, api error InsufficientInstanceCapacity: We currently do not
  have sufficient i8g.xlarge capacity in the Availability Zone you requested
  (us-east-1c). ... You can currently get i8g.xlarge capacity by not specifying
  an Availability Zone in your request or choosing us-east-1a, us-east-1b,
  us-east-1d.
  Build 'amazon-ebs.aws' errored after 2 minutes 14 seconds: ...
```

Fix: wrap the packer invocation in a retry loop (MAX_ATTEMPTS=3). Because subnet_filter uses "random": true, each attempt re-picks a subnet and thus likely a different AZ. Only InsufficientInstanceCapacity is retried (with a RETRY_SLEEP=90s pause); a successful build breaks out immediately, and any other failure (or exhausted capacity retries) exits at once, preserving packer's exit code. This will allow not to rerun all the job manually again

- [X] verification: https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/124/

### DIFF
--- a/packer/build_image.sh
+++ b/packer/build_image.sh
@@ -377,31 +377,46 @@ mkdir -p build
 export PACKER_LOG=1
 export PACKER_LOG_PATH
 
-set -x
-/usr/bin/packer ${PACKER_SUB_CMD} \
-  -only="$TARGET" \
-  -var-file="$JSON_FILE" \
-  ${EXTRA_VAR_FILE_ARG:+"$EXTRA_VAR_FILE_ARG"} \
-  -var install_args="$INSTALL_ARGS" \
-  -var ssh_username="$SSH_USERNAME" \
-  -var scylla_full_version="$SCYLLA_FULL_VERSION" \
-  -var scylla_version="$VERSION" \
-  -var scylla_machine_image_version="$SCYLLA_MACHINE_IMAGE_VERSION" \
-  -var scylla_python3_version="$SCYLLA_FULL_VERSION" \
-  -var creation_timestamp="$CREATION_TIMESTAMP" \
-  -var scylla_build_sha_id="$SCYLLA_BUILD_SHA_ID" \
-  -var build_tag="$BUILD_TAG" \
-  -var environment="$ENV_TAG" \
-  -var operating_system="$OPERATING_SYSTEM" \
-  -var branch="$BRANCH" \
-  -var ami_regions="$AMI_REGIONS" \
-  -var arch="$ARCH" \
-  -var product="$PRODUCT" \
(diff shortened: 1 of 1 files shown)


yaronkaikov (approved):
LGTM

dimakr:
Currently random selection has no memory. Is there a chance that 3 attempts would hit the exhausted AZ (or AZs)?
@Annamikhlin

Annamikhlin:
> Currently random selection has no memory. Is there a chance that 3 attempts would hit the exhausted AZ (or AZs)?

@dimakr Yes, it's memoryless. so it’s still possible to hit the same unavailable AZ again, this reduces the risk but doesn’t eliminate it. In our case there are 4 possible AZs and only 1 had no capacity, so each retry has a good chance of picking a healthy one. With 3 retries, the odds are about 98% that at least one will succeed. It mainly becomes a problem only if multiple AZs are out of capacity at the same time, which is uncommon, and the 90s wait between retries gives AWS a chance to free up resources.

fruch:
@Annamikhlin why no using the internal feature https://github.com/hashicorp/packer/pull/9061

> > Currently random selection has no memory. Is there a chance that 3 attempts would hit the exhausted AZ (or AZs)?
> 
> @dimakr Yes, it's memoryless. so it’s still possible to hit the same unavailable AZ again, this reduces the risk but doesn’t eliminate it. In our case there are 4 possible AZs and only 1 had no capacity, so each retry has a good chance of picking a healthy one. With 3 retries, the odds are about 98% that at least one will succeed. It mainly becomes a problem only if multiple AZs are out of capacity at the same time, which is uncommon, and the 90s wait between retries gives AWS a chance to free up resources.

90s wouldn't get you resource frees
and random isn't the best

but it's better then current situation

I'm wondering if we can built the image of other more available arm base instance types,
that worth trying out

fruch (approved):

PR #987, opened 2026-07-14
fix(oci): network bandwidth estimation wrong for flex instances

fruch:
code was continuing and overwriting the OCPU base computation needed for flex instances, which ended up with too small streaming limit that was slowing down backup/restore operations

Fixes: SMI-300


vponomaryov (approved):
LGTM

PR #991, opened 2026-07-23
improvement(oci): expose image ID as a variable

vponomaryov:
It will allow easier switch of a base region we build OCI image in.

Ref: #SCT-651

### DIFF
--- a/packer/oci_variables.json
+++ b/packer/oci_variables.json
@@ -1,6 +1,7 @@
 {
   "oci_availability_domain": "",
   "oci_subnet_ocid": "",
+  "oci_base_image_ocid": "",
   "oci_instance_shape": "VM.Standard3.Flex",
   "oci_instance_shape_config_ocpus": "2",
   "oci_instance_shape_config_memory_in_gbs": "16"
--- a/packer/scylla.json
+++ b/packer/scylla.json
@@ -193,7 +193,7 @@
       "availability_domain": "{{user `oci_availability_domain`}}",
       "compartment_ocid": "{{user `oci_compartment_ocid`}}",
       "subnet_ocid": "{{user `oci_subnet_ocid`}}",
-      "base_image_ocid": "ocid1.image.oc1.iad.aaaaaaaan5hami2nl7girpphwwfwz33iawrfthuks2x3qow5a2lmozx7gt2a",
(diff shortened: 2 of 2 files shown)


packer/oci_variables.json
```diff
@@ -1,6 +1,7 @@
 {
   "oci_availability_domain": "",
   "oci_subnet_ocid": "",
+  "oci_base_image_ocid": "",
```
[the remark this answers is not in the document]
vponomaryov: This is expected limitation.

Must be merged only in parallel to the following PR:
- https://github.com/scylladb/scylla-pkg/pull/6492

And update of the jenkins creds with the image from the eu-phoenix-1 region

packer/scylla.json
```diff
@@ -193,7 +193,7 @@
       "availability_domain": "{{user `oci_availability_domain`}}",
       "compartment_ocid": "{{user `oci_compartment_ocid`}}",
       "subnet_ocid": "{{user `oci_subnet_ocid`}}",
-      "base_image_ocid": "ocid1.image.oc1.iad.aaaaaaaan5hami2nl7girpphwwfwz33iawrfthuks2x3qow5a2lmozx7gt2a",
+      "base_image_ocid": "{{user `oci_base_image_ocid`}}",
```
yaronkaikov, line 196: who will provide now the base_image_ocid ?

vponomaryov: I wrote it in the comment answering AI review:
- https://github.com/scylladb/scylla-machine-image/pull/991#discussion_r3637763685

Jenkins creds must be updated with the proper image.

yaronkaikov: I don't understand how Jenkins has anything to do with that. as far as i understand, you want to run artifact on different region, right? so in oci-image pipeline, you do a copy image, get the id and provide it for the test. is that correct?

vponomaryov: See https://github.com/scylladb/scylla-pkg/pull/6492#issuecomment-5058170299

fruch: let's do it in two steps
1) update jenkins with this existing image paramter, and have this merged 
2) use a new cred for the new region parameters, and use it for testing the scylla-pkg changers  

and that secret should be region specific from now on

fruch (approved):

PR #1002, opened 2026-08-17
gcp: sync VPC MTU on boot and warn when jumbo frames are not enabled

juliayakovlev:
Query the VPC-configured MTU from GCP metadata on every boot and apply it to eth0 as a defensive reconcile. When the VPC MTU is below 8896 (GCP's maximum / jumbo frames), log a warning in the journal and display it at login with the exact gcloud command to fix it.

- Add GcpInstance.network_mtu property and gcp_mtu_warning() helper
- Add sync_gcp_mtu_from_vpc() to scylla_image_setup (runs every boot)
- Implement GcpInstance.check() for login banner warning
- Drop @staticmethod from CloudInstance.check() base declaration
- Add tests for network_mtu and check() behavior
- Document VPC MTU requirement in SETUP.md

Refs: SMI-248

PR #1005, opened 2026-08-26
gcp: sync VPC MTU on boot and warn when jumbo frames are not enabled

juliayakovlev:
Query the VPC-configured MTU from GCP metadata on every boot and apply it to eth0 as a defensive reconcile. When the VPC MTU is below 8896 (GCP's maximum / jumbo frames), log a warning in the journal and display it at login with the exact gcloud command to fix it.

- Add GcpInstance.network_mtu property and gcp_mtu_warning() helper
- Add sync_gcp_mtu_from_vpc() to scylla_image_setup (runs every boot)
- Implement GcpInstance.check() for login banner warning
- Drop @staticmethod from CloudInstance.check() base declaration
- Add tests for network_mtu and check() behavior
- Document VPC MTU requirement in SETUP.md

The non-mapping user-data guard in `lib/user_data.py` is now its own commit, per @fruch's review feedback on the previous PR. It is a pre-existing bug, but the MTU work depends on it: `GcpInstance.tier1_override` adds a new boot-path read of user-data, and without the guard a non-mapping user-data raises `AttributeError: 'str' object has no attribute 'get'`, which `sync_gcp_mtu_from_vpc()` swallows — silently skipping MTU configuration.

Supersedes #1002, which was opened from a branch in the scylladb repo by mistake; this one comes from my fork. Review history is in #1002.

Testing:

- [x] [create an image and run artifact test and GCE test](https://jenkins.scylladb.com/job/releng-testing/job/next-machine-image/133/)

Refs: SMI-248

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -364,6 +364,28 @@ uv venv                    # Create virtual environment
 5. Run integration tests: `make test-integration` (requires AWS)
 6. Start developing!
 
+## GCP: Enabling Jumbo Frames for Tier 1 Instances
+
+By default, GCP VPCs use an MTU of 1460. For Tier 1 networking instances,
+the VPC should be configured with jumbo frames (MTU 8896) to achieve full
+streaming throughput. The guest image cannot raise the MTU beyond what the
+VPC allows, so this is a VPC-level requirement.
+
+### Steps
+
+1. Set the VPC MTU to 8896:
+   ```bash
+   gcloud compute networks update <network> --mtu=8896
+   ```
+
+2. **Stop and start** (not reboot) all instances on that network. A guest
+   reboot does not refresh the MTU — only a stop/start causes GCP to
+   re-provision the NIC with the new VPC MTU via DHCP.
+
+The Scylla image will log a warning at login and in the journal when a
+Tier 1 instance has a VPC MTU below 8896, including the exact `gcloud`
+command to fix it. Standard (non-Tier 1) instances are not affected.
+
 ## Resources
 
 - [UV Documentation](https://github.com/astral-sh/uv)
--- a/common/scylla_image_setup
+++ b/common/scylla_image_setup
@@ -8,14 +8,131 @@ import logging
 import os
 import re
 import subprocess
+import urllib.error
 from pathlib import Path
 from subprocess import run
 
 from lib.log import setup_logging
-from lib.scylla_cloud import get_cloud_instance, is_azure, is_ec2
+from lib.scylla_cloud import (
+    GCP_MAX_MTU,
+    GCP_MTU_WARNING_PATH,
+    GCP_STANDARD_MTU,
+    gcp_mtu_warning,
+    get_cloud_instance,
+    is_azure,
+    is_ec2,
+    is_gce,
+)
 
 
 LOGGER = logging.getLogger(__name__)
+
+
+NETPLAN_MTU_PATH = Path("/etc/netplan/99-scylla-mtu.yaml")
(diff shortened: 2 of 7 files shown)


juliayakovlev:
## Testing performed

### Unit tests
`pytest tests/` — 301 passed. `ruff check` and `ruff format --check` clean.

### Manual validation on GCE

Image `debug-scylla-2026-4-0-dev-x86-64-2026-08-26t17-43-00` built from this branch,
`n2-standard-48` (smallest Tier-1-capable `n2-standard` in `gcp_net_params.json`),
default VPC (MTU 1460), Tier 1 asserted via `--metadata=scylla_tier1_networking=true`.

**1. Happy path — setup succeeds** (8 × NVMe local SSD)

- `scylla-image-setup`: `ActiveState=active`, `Result=success`, `ExecMainStatus=0`
- `/dev/md0` mounted at `/var/lib/scylla`, `machine_image_configured` created
- detection: `network_mtu=1460`, `tier1_override=None`, `is_tier1_networking=True`
- journal: `Tier 1 instance detected. Setting MTU on eth0 to 1460 (from VPC metadata)`, then the `WARNING`
- `/etc/scylla/gcp_mtu_warning` written 56s after boot, mode `644`
- `eth0` MTU 1460; `/etc/netplan/99-scylla-mtu.yaml` absent (Tier 1 clears the clamp)
- login banner prints the warning in yellow

**2. Failure path — what the third commit fixes** (no local SSD, so `scylla_create_devices` aborts)

Reproduced on an image built before that commit: `scylla_configure.py` exits 1 and the
script dies before reaching `sync_gcp_mtu_from_vpc()` — no MTU applied, no warning
recorded, on a boot where the operator most needs it.

With the `finally`, the journal shows the sync running *after* the abort:

```
15:00:23  scylla_image_setup[1891]: No data disk found, abort setup
15:00:23  scylla_image_setup[1086]: Tier 1 instance detected. Setting MTU on eth0 to 1460
15:00:23  scylla_image_setup[1086]: WARNING - VPC MTU is 1460; jumbo frames are not enabled...
```

The original failure is not masked: `ActiveState=failed`, `Result=exit-code`,
`ExecMainStatus=1`, and `machine_image_configured` stays absent so setup retries on
the next boot.

**3. Runs on every boot, not just the first**

With `machine_image_configured` present, removed the warning file and restarted the
unit: the guarded block was skipped (0 occurrences of `Going to create scylla.yaml`)
while the MTU sync still ran and recreated `/etc/scylla/gcp_mtu_warning`.

### Not covered

- **Warning-clearing transition** (raise VPC MTU to 8896, stop/start): needs a dedicated
  VPC — changing the shared project's `default` network MTU affects every instance on it.
- **Standard (non-Tier-1) clamp path** on a real instance: unit-tested only.
- **CI does not exercise this feature at all.** The GCE artifact test runs on
  `n1-standard-2`, which is absent from `gcp_net_params.json` (the file has no `n1-*`
  entries), so `is_tier1_networking` returns `None` and `sync_gcp_mtu_from_vpc()`
  deliberately does nothing. Automated coverage would need either `n1` entries added
  (which changes behaviour for those types) or the test instance type changed in
  scylla-pkg — neither belongs in this PR.
- The `finally` behaviour itself has **no unit test**: it lives in the
  `if __name__ == "__main__":` block, which is not importable. It was verified via the
  boot described in (2), plus a local `runpy` harness driving the block with
  `scylla_configure.py` mocked to exit 1.

All test instances have been deleted.

juliayakovlev:
@fruch please, review

fruch (approved):

PR #1007, opened 2026-08-27
packer: replace `vim-nox` with `vim` to drop Ruby from cloud images

vponomaryov:
vim-nox is Vim built with every embedded scripting interpreter,
which is the sole reason Ruby is installed in our cloud images.
Plain vim is the same editor without those interpreters,
so the whole stack drops out and the image gets smaller.

Also drop the redundant, misspelled `vim.tiny` argument.

Ref: https://github.com/scylladb/siren-devops/issues/1764

Closes: #SMI-250

### DIFF
--- a/packer/scylla_install_image
+++ b/packer/scylla_install_image
@@ -257,7 +257,7 @@ if __name__ == "__main__":
     apt_get(
         f"install -y --auto-remove {args.product}-machine-image {args.product}-server-dbg "
         "chrony cpufrequtils curl dnsutils ethtool htop initramfs-tools jq mdadm ncat netcat-openbsd "
-        "net-tools nftables nload nmap python3-boto3 sysstat systemd-coredump tmux traceroute vim-nox vim.tiny wget xfsprogs aide"
+        "net-tools nftables nload nmap python3-boto3 sysstat systemd-coredump tmux traceroute vim wget xfsprogs aide"
     )
     apt_get(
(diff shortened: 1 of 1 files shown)


zgrim (approved):
Nice catch @vponomaryov 
vim still uses python3, but i guess that interpreter is more fundamental to the system and is installed anyway, no point in avoiding it

btw `vim.tiny` was misspelled but actually got installed, apt resolves that dot as a regex and matched `vim-tiny`, so we installed 2 versions of vim, the full one (from nox) with all interpreters as /usr/bin/vim and also the minimal one as /usr/bin/vi

fruch:
> Nice catch @vponomaryov vim still uses python3, but i guess that interpreter is more fundamental to the system and is installed anyway, no point in avoiding it
> 
> btw `vim.tiny` was misspelled but actually got installed, apt resolves that dot as a regex and matched `vim-tiny`, so we installed 2 versions of vim, the full one (from nox) with all interpreters as /usr/bin/vim and also the minimal one as /usr/bin/vi

it's actually @mykaul that observed it, and we just following it along.

we just need you confirmation that this nox variant isn't really needed, and a vanilla vim package would be enough.
if we get it wrong, it's a long route to fix it.

zgrim:
> we just need you confirmation that this nox variant isn't really needed

cloud backend has no use for it, i presume it's there for cx/devops

PR #1017, opened 2026-09-08
docs/plans: plan for refreshing io_properties during upgrade

fruch:
Design proposal for [SMI-218](https://scylladb.atlassian.net/browse/SMI-218) — for review before any implementation lands. Nothing here changes behaviour; it adds one document under `docs/plans/`.

## The problem

`io_properties.yaml` is written once, on the first boot of an instance, and is never revisited. Every correction we make to the I/O parameter tables afterwards is invisible to fleets already running — i7ie, i7i and i8g/i8ge values have all changed since they started shipping. The only remedy today is running `scylla_io_setup` by hand on every node, which means minutes of `iotune` per node, the very thing the preset tables exist to avoid.

The same blind spot means a stop → resize → start leaves the file describing the previous instance type, silently.

## What the plan proposes

Treat the file as what it actually is — a derived artifact of *(parameter table, cloud, instance type, disk count)* — and reconcile it on upgrade, reboot and resize, instead of freezing it at first boot. The rolling upgrade already stops and starts Scylla, so a corrected file inside that window costs the user nothing.

Four building blocks:

1. **Record provenance** — so we know which files are ours to update.
2. **Keep the history of published values** — so nodes that predate that record can still be recognised as ours.
3. **Classify ownership before acting** — update what we own, leave everything else untouched with a `.new` candidate and one copy-paste command.
4. **Three triggers, one idempotent operation** — every boot, package upgrade, and on demand.

## Targets

- Zero manual steps on upgrade and on reboot/resize for files we can prove we wrote.
- Never silently clobber a hand-tuned file.
- Refresh is a table lookup — under a second, never `iotune`.
- Every automatic change auditable and revertible with one command.
- All four clouds, both packaging formats.

Headline metric: on a fleet upgraded from an image predating the i7ie re-measurement, every i7ie node runs current values after the standard rolling upgrade, with no step added to the upgrade documentation.

## Cross-check with prior art

This is not a novel problem, so the plan follows the shape of the established solutions and says so explicitly: `ucf(1)`'s ownership model (including its record of previously published versions, which is exactly the legacy-instance problem), dpkg/rpm conffile semantics, Ceph's mclock sanity bounds on automatic performance-affecting values, the vendor-data-plus-trigger pattern from `ca-certificates`/`tuned`, and the reconcile-on-change approach scylla-operator already takes for node tuning.

## Open questions for reviewers

- Should nodes holding likely-`iotune` output default to *notify* (safer) or *auto* (fixes more of the SMI-218 fleet without user action)? The plan proposes deciding this per instance family through a reviewed override rather than globally.
- Where the policy setting lives — machine-image config, user-data schema, or both.
- Which maintained branches need this backported for the i7ie fleet.

Both positions from the ticket are kept: the "generate it and let the user decide" idea, narrowed to the genuinely ambiguous cases, and the "generating takes too long" objection, answered because the refresh path is a lookup and never runs `iotune`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[SMI-218]: https://scylladb.atlassian.net/browse/SMI-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/plans/io-properties-upgrade-refresh.md
+++ b/docs/plans/io-properties-upgrade-refresh.md
@@ -1,0 +1,219 @@
+# Refreshing `io_properties.yaml` During Upgrade
+
+Tracking issue: [SMI-218](https://scylladb.atlassian.net/browse/SMI-218)
+(migrated from [scylla-machine-image#748](https://github.com/scylladb/scylla-machine-image/issues/748))
+
+## 1. Problem Statement
+
+`/etc/scylla.d/io_properties.yaml` is written exactly once, on the first boot of an
+instance, and is never revisited. Every later correction we make to the I/O parameter
+tables is therefore invisible to already-running fleets.
+
+What this costs us today:
+
+- **Corrected values never reach existing nodes.** We have re-measured and changed
+  values for families that were already shipping — i7ie, i7i, i8g/i8ge — and in some
+  cases changed the shape of the lookup keys. A node launched before those changes keeps
+  the old numbers for the rest of its life, even after `scylla-machine-image` is upgraded
+  to a version that knows better.
+- **The families SMI-218 names are the worst case.** i7ie was not a supported class at
+  all in earlier releases, so we never generated a file for it: the login banner told the
+  user to run `scylla_io_setup` themselves. Those nodes hold single-run `iotune` output
+  rather than our vetted values.
+- **The only remedy is slow and manual.** Running `scylla_io_setup` per node takes
+  minutes of `iotune` time — the very reason the preset tables exist — and has to be
+  driven node by node by the user.
+- **Instance resize is silently wrong.** Stop instance → change type → start, and the
+  file still describes the previous instance type. Nothing detects this.
+- **The upgrade already gives us a free restart.** Scylla reads the I/O properties only
+  at startup, and the documented rolling upgrade already stops Scylla, upgrades packages,
+  and starts it again. Correcting the file inside that window costs the user nothing.
+
+**The aim of this plan: correct I/O properties land as part of the ordinary upgrade
+sequence, with no added instruction for the common case, and without ever silently
+discarding a file the user tuned themselves.**
+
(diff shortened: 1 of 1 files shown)


docs/plans/io-properties-upgrade-refresh.md
```diff
@@ -0,0 +1,219 @@
...
+
+## 5. Testing Requirements
+
+- **Unit:** value computation per cloud (including disk-count scaling and family-wide
+  fallback keys); the full ownership matrix, with a real pre-change i7ie file as a
+  fixture; history matching; implausible-change rejection; backup rotation and rollback.
+- **Packaging:** post-install hook is idempotent, is a no-op without cloud metadata or
+  mounted storage, and never fails the transaction; fresh install → upgrade → downgrade
+  on both Debian and RPM.
+- **Integration, per cloud:** fresh boot unchanged (regression); a node launched from an
```
fruch, line 184: we should run the relevant images rolling upgrades in SCT

roydahan:
The plan isn’t very detailed and a bit ambiguous.
It can be shorter and to the point.

IIUC the open question part, so we need a way to “auto” check if there are updates to the io.file and update it if the instance type being used have any relevant changes.

we should also have a flag or a way to not update anything (e.g. one did a manual run of iotune and want to keep the values as is).

fruch:
> The plan isn’t very detailed and a bit ambiguous. It can be shorter and to the point.
> 
> IIUC the open question part, so we need a way to “auto” check if there are updates to the io.file and update it if the instance type being used have any relevant changes.
> 
> we should also have a flag or a way to not update anything (e.g. one did a manual run of iotune and want to keep the values as is).

it does suggest how to "sign" the file in a way that can be detected, so we know if a provided file or was manually altered

